PR #26, opened 2020-10-03
Update README.md

sofia100:
added the attractive YouTube thumbnail instead of the only link text


Nikhil-Nandagopal:
@sofia100 please refrain from making changes without raising an issue and discussing with a maintainer. This is invalid because the embed tag will add a complete youtube player instead of just the image. Please read our contribution document before contributing.

PR #32, opened 2020-10-03
#858 gif for table widget

gogetter22:


widget-reference/table.md
```diff
@@ -63,3 +63,4 @@ Tables are useful to view large lists of data. To drill down into the data of a
 | **onRowSelected** | Sets the action to be run when user selects a row in the table. Default supported actions are: Call API, Execute DB Query, Navigate to Page, Show Alert, Open / Close popup |
```
Nikhil-Nandagopal, line 63: @gogetter22 can you update the gif to drop the table, resize it so it's more visible, and bind it to a query.

Nikhil-Nandagopal (changes requested):
@gogetter22 requested changes

gogetter22:
@Nikhil-Nandagopal : Made the changes, please review.

widget-reference/table.md
Nikhil-Nandagopal: @gogetter22 something seems wrong with the table in the gif, the bg is transparent instead of white. How did this happen?

Nikhil-Nandagopal (changes requested):
- The Table in the Gif is looking strange
- The table is calling Queries instead of displaying data from a query
- Create a `select * from users` query and run it, bind the query to the table data property as {{ query.data}}

PR #39, opened 2020-10-06
Docs/improve navigate to

visibleajay:


function-reference/navigateTo.md
```diff
@@ -15,7 +15,8 @@ navigateTo(pageName: string) -> void
 
 | **Argument Name** | **Description** |
 | :--- | :--- |
-| **pageName** |  Page name to which we would like to be transported. It's case sensitive. |
+| **pageNameOrUrl** |  Page name or URL to which we would like to be transported. PageName is case sensitive. |
+| **params** |  Query params, to share information with destination. |
```
mohanarpit, line 19:
```suggestion
| **params** |  Query parameters passed via the URL. Used to share information with destination page. |
```

mohanarpit:
@visibleajay Can you also please add this page to the `Summary.md` file? Without that, it'll not show up in the documentation section at docs.appsmith.com

visibleajay:
@mohanarpit, @Nikhil-Nandagopal has already made requisite the [Change](https://github.com/appsmithorg/appsmith-docs/commit/a9f4019361bd239fa23097c4e59e7402c38ebd71#diff-b4f3573b2f24d5af026c33acf52ff716) and it's already visible at [navigateTo Docs](https://docs.appsmith.com/function-reference/navigateto). Thank You!.

mohanarpit (dismissed):

function-reference/navigateTo.md
mohanarpit: Sorry, I just noticed. Since we are adding the `params` to the Arguments section here, can we also update the `Signature` section to reflect this? Else, the arguments is out of sync with the function signature.

function-reference/navigateTo.md
```diff
@@ -15,7 +15,7 @@ navigateTo(pageName: string) -> void
 
 | **Argument Name** | **Description** |
 | :--- | :--- |
-| **pageName** |  Page name to which we would like to be transported. It's case sensitive. |
+| **pageNameOrUrl** |  Page name or URL to which we would like to be transported. PageName is case sensitive. |
+| **params** |  Query parameters passed via the URL. Used to share information with destination page. |
```
mohanarpit, line 19: Sorry, I just noticed. Since we are adding the `params` to the Arguments section here, can we also update the `Signature` section to reflect this? Else, the arguments is out of sync with the function signature.

visibleajay: Apologies, I didn't noticed it. The `params` are optional, should we use `navigateTo(pageName: string, params?: {}) -> void` or a normal `navigateTo(pageName: string, params: {}) -> void` would be better.

[navigateTo Signature](https://github.com/appsmithorg/appsmith/blob/master/app/client/src/utils/autocomplete/EntityDefinitions.ts#L228)

Nikhil-Nandagopal: @visibleajay we can use params?: {} and mention in the argument table that it is optional

mohanarpit (approved):

PR #42, opened 2020-10-09
Added Image In Demo Video In README

KaranSinghBisht:
So the README.md was looking pretty bad so I added this 
I have used an imgur dummy image which can be easily replaced with a permanent one
This is how the README.md looks like now
The image is clickable and redirects to the youtube video, sorry for the bad quality ss
Screenshot 2020-10-10 032118

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,5 +25,5 @@ Appsmith makes building internal tools as easy as
 
 Check out our demo video to see how easy it is!
 
-{% embed url="https://www.youtube.com/watch?v=mzqK0QIZRLs&feature=youtu.be" %}
-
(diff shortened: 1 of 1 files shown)


mohanarpit:
@KaranSinghBisht Thank you for the PR. Unfortunately, this is not valid because the embed URL is expected. This is rendered properly at [docs.appsmith.com](https://docs.appsmith.com). There you will see the demo video showing up properly.

KaranSinghBisht:
Oh ok, I opened the pr because it was looking really bad at this repo

Can you remove the invalid tag as this pr was not really spammy

As per Hacktoberfest guidelines this pr shouldn't be marked invalid

Plus invalid label might lead to disqualification...

Please use the invalid label wisely @mohanarpit

PR #48, opened 2020-10-19
Add button recaptcha documentation

JoshHeng:
Fixes [#1268](https://github.com/appsmithorg/appsmith/issues/1268) in appsmithorg/appsmith

May need updating if the Recaptcha integration is updated with a server side check


widget-reference/button.md
```diff
@@ -12,10 +12,16 @@ description: >-
 | **Button Style** | Sets the style of the button. You can choose from three button styles: Primary, Secondary & Danger Button. |
 | **Visible** | Controls widget's visibility on the page. When turned off, the widget will not be visible when the app is published  |
```
Nikhil-Nandagopal, line 13: @JoshHeng the gif doesn't show the user using re-captcha and sending the token to the server. Can we make this a separate Page with clear instructions to create a recaptcha site key, configure the button and send the token to the backend. Also need information on how the server will verify the token

JoshHeng: I've separated the integration instructions to a separate page. 
I don't think the button gif should show adding the recaptcha key/verification as I think it should just show the client side/appsmith configuration of the button

Nikhil-Nandagopal: @JoshHeng I think there was some confusion here. I mean, can we have a separate gif not change the button gif. Which shows a user adding the key to the button, creating an API in the API pane and binding the buttons token to the API body i,e {{ Button1.token }}

widget-reference/button.md
```diff
@@ -12,10 +12,10 @@ description: >-
...
 | **Disabled** | Disables input to the widget. The widget will remain visible to the user but a user input will not be allowed.  |
+| **Google Recaptcha Key** | Adds a [Google ReCAPTCHA v3](https://www.google.com/recaptcha/) check to the button (see [Google reCAPTCHA](../third-party-services/google-recaptcha.md)) |
```
Nikhil-Nandagopal, line 15: An internal property for the stored re-captcha token needs to be created
recaptchaToken is the key

JoshHeng: I've added instructions for the API pane

Nikhil-Nandagopal (approved):

Nikhil-Nandagopal:
@JoshHeng thank you for the contribution! Sorry we had to go back and forth a little on it but I think it came out really well 😃

JoshHeng:
Thanks for all the help!

PR #61, opened 2020-10-30
Bullet list of form widgets

valerianpereira:
Bullet list of form widgets - Simple and standard doc across like the other pages.


Nikhil-Nandagopal:
@valerianpereira this is a low-quality PR and doesn't add much value to our documentation as such I'm rejecting it. If you would like to contribute in a meaningful way, I suggest having a look at our many open good first issues 😃

PR #73, opened 2020-12-31
Update getting-started.md

Sayanta66:
Added a description on what Appsmith does


hetunandu:
Hey @Sayanta66 Thanks for this PR but this is not a good quality contribution. I would suggest you follow our contribution guides and pick up one of the `Good First Issues` to get started.

PR #84, opened 2021-01-29
Added 4 db plugin docs

nidhi-nair:


datasource-reference/querying-dynamodb.md
```diff
@@ -0,0 +1,34 @@
...
+### **Connection**
+
+You need to fill in the following parameters:
+
+* **Region\*:** Select the region where your DynamoDB instance exists
+* **Host Address / Port\*:** Fill in the database host’s address and port
+* **AWS Access Key Id\*:** The identifier for your AWS Access key that is configured to have access to this database
+* **AWS Secret Access Key\*:** The secret key for this identifier 
+
+## Querying DynamoDB
```
mohanarpit, line 24: Can we enhance this documentation with examples & screenshots of how to query DynamoDB using Appsmith? Check out the Firestore docs as inspiration.

nidhi-nair: @mohanarpit Should I limit the queries in the docs to the basic CRUD ones or should I link to a separate page for each query and cover them all? There's 40+ methods that DynamoDB supports

mohanarpit: We don't need to cover all query types. Just specifying the Create & Read queries will go a long way in getting people to understand how this is done. For the rest, we just link to the DynamoDB documentation itself to point people towards all the other query types. We don't need to replicate DynamoDB's docs.

PR #128, opened 2021-03-24
Update store-value.md

hetunandu:
Updates to the store value function
Implementation PR: appsmithorg/appsmith#3663

### DIFF
--- a/framework-reference/store-value.md
+++ b/framework-reference/store-value.md
@@ -1,5 +1,5 @@
 ---
-description: This function is used to store key-value data within browser local storage.
+description: This function is used to store key-value data to access later anywhere in the app.
 ---
 
 # Store Value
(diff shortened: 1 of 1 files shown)


framework-reference/store-value.md
```diff
@@ -18,6 +18,7 @@ storeValue(key: string, value: any): void
 | :--- | :--- |
 | **key** | Name of the key to store the value against |
 | **value** | Value to give the key you are creating/updating |
+| **persist**| Should this key value get persisted in the browser local storage to use between sessions. Defaults to true |
```
Nikhil-Nandagopal, line 21: @hetunandu shouldn't we default to false? Can we migrate older apps to a deprecated method or something?

hetunandu: Migration would require some script parsing which we can't do today.

hetunandu: As discussed, we would tackle this problem for future releases

Nikhil-Nandagopal (approved):

PR #132, opened 2021-04-06
Update sendgrid.md

miloparsons:
Changed value for APPSMITH_MAIL_USERNAME to correct value (apikey)


Nikhil-Nandagopal (approved):

Nikhil-Nandagopal:
@miloparsons thank you for the contribution! I did not know that apikey was the standard name for all Sendgrid configurations!

PR #135, opened 2021-04-30
update s3 download guide.

sumitsum:
- since the download settings in property pane is not working, replacing the guide for now with mustache expression flow.
- adding more images and gif.
- add how to get count of mongo documents without fetching them.


sumitsum:
Closing it in favor of new PR.

PR #143, opened 2021-05-11
fix: remove uploadFileUrls from property reference

Tooluloope:
- Update file widget, removed uploadFileUrls

### DIFF
--- a/widget-reference/filepicker.md
+++ b/widget-reference/filepicker.md
@@ -26,7 +26,6 @@ See our guides on
 | Internal Property | Description |
 | :--- | :--- |
 | **files** | This is the array of files selected in the file picker. The data of each file are present in each array object and can be accessed using  |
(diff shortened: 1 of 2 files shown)


widget-reference/filepicker.md
```diff
@@ -26,7 +26,6 @@ See our guides on
 | Internal Property | Description |
 | :--- | :--- |
 | **files** | This is the array of files selected in the file picker. The data of each file are present in each array object and can be accessed using  |
-| **uploadedFileUrls** | This is the data that was bound in the Uploaded File URLs property for access to be uploaded in an API. |
```
somangshu, line 29: There is another occurrence on line number 43, Remove that as well

somangshu (changes requested):

widget-reference/filepicker.md
Tooluloope: Removed

somangshu (approved):

PR #157, opened 2021-06-09
Adding documentation for Mongo Form

trishaanand:


datasource-reference/querying-mongodb/README.md
```diff
@@ -64,7 +64,185 @@ The SSL Mode can be set to one of the following values:
...
+
+For an easy to use interface to query your mongo database, choose `Form input`.
+
+As part of Form Input, Appsmith supports queries like `Find one or more documents`, `Insert a document`, `Update one`, `Count` etc.
+
+All mongo queries return an array of objects where each object is a mongo document and properties of the object are the keys of the document.
+
+![](../../.gitbook/assets/mongo-form.gif)
+
+### 1. Find one or more documents
```
Nikhil-Nandagopal, line 77: A user yesterday had no idea what the difference between update one and update one or many was. Why do we need this and why is it so complex?
Can it not just be
Find Document(s)
Insert Document(s)
Update Documents(s) (add a limit)
Delete Document(s)

The UI is too verbose right now

trishaanand: Ok. Let me add a limit to update and then change the fields required for either of the two commands for update.

trishaanand: I have removed a lot of fields that mongo db supports and kept it to the ones that I personally have used in the past two years to ensure that generic daily use cases are easily covered as part of form and the user would not have to move to RAW query format which is harder to grok and understand.

Nikhil-Nandagopal: Cool can we update the list and the docs, I'll approve it after

Nikhil-Nandagopal (approved):

PR #165, opened 2021-07-01
Update embed-appsmith-into-existing-application.md

jmcouffin:
added info on how to make the app in the iframe fullscreen


vihar (approved):
Thanks, @jmcouffin! This is helpful.

PR #191, opened 2021-09-01
Fix header spelling

thedatadavis:


vihar (approved):
That's a nice catch, thanks @thedatadavis!

PR #204, opened 2021-09-27
feat: Migrating to new docker image (single container)

geekup-legodevops:
-- Document guides how to migrate data & configuration from Docker stack to new appsmith docker image (single container)


setting-up/migrate.md
```diff
@@ -0,0 +1,125 @@
...
+APPSMITH_MONGODB_URI=mongodb://<MONGO_USERNAME>:<MONGO_PASSWORD>@localhost/appsmith
+APPSMITH_REDIS_URL=redis://127.0.0.1:6379
+```
+### 3. Export https config & certificate (Optional)
+
+* If you have deployed Appsmith with custom domain, you might want to move your certificate to the new container
+
+```
+# Add custom domain to configuration file
+echo APPSMITH_CUSTOM_DOMAIN=<YOUR-DOMAIN> >> $NEW_APPLICATION_DIR/stacks/configuration/docker.env
```
sharat87, line 86: If the user already has a custom domain configured, like the above paragraph says, then the `docker.env` would already have this env variable set right? So do we still need this command to be run here?

setting-up/migrate.md
```diff
@@ -0,0 +1,125 @@
...
+sudo cp -rf $OLD_APPLICATION_DIR/data/certbot/conf/* $NEW_APPLICATION_DIR/stacks/letsencrypt
+```
+
+### 4. Setup new Appsmith with Fat container
+* Before you can start new appsmith docker, please shutdown current appsmith application first
+```
+cd $OLD_APPLICATION_DIR
+docker-compose down
+```
+* Now, you can start to set up Appsmith with new Docker image. Please follow this guideline to start up a new container: [https://github.com/appsmithorg/appsmith/blob/release/deploy/docker/README.md](https://github.com/appsmithorg/appsmith/blob/release/deploy/docker/README.md)
```
sharat87, line 101: Can we instead link to <https://docs.appsmith.com/setup/docker#docker-compose-configuration> here, if that works just as well?

setting-up/migrate.md
```diff
@@ -0,0 +1,125 @@
...
+
+* After successfully starting up new container, you can import your previous data by exported file from the above step
+* By using `Appsmith CLI`, new feature we have nearly introduced in the new Docker image, you can import all data with only one command
+
+```
+# Create folder restore
+mkdir -p $NEW_APPLICATION_DIR/stacks/data/restore
+
+# Copy exported file 
+sudo cp $OLD_APPLICATION_DIR/data/mongo/db/backup/appsmith-data.archive $NEW_APPLICATION_DIR/stacks/data/restore/appsmith-data.archive
```
sharat87, line 114: I see that we are not including `sudo` in the docker/docker-compose commands, which makes me wonder if we should included here at all? Can we remove it?

sharat87:
Note to self: Add instructions on how to get a superadmin in the instance, after migration is finished.

PR #226, opened 2021-10-02
docs: Updated the links for GMap + Telemetry

UBG100:
The Heroku page didn’t have updated links for the Google Maps and Telemetry widgets, which just led you to 404 pages.


UBG100:
In case the reviewers are confused, this was made before v1.3, where the links were broken in v1.2.1 "customize default settings in Heroku". It seems as if a large section was removed in v1.3

UBG100:
This PR was made before v1.3

Issue #149, opened 2021-05-25
Update/Edit Appsmith Heroku Setup


vihar:
The current documentation for the Appsmith `Heroku` setup needs a lot of editing and formatting. The images aren't also clear.

Link: https://docs.appsmith.com/setup/heroku

PR #236, opened 2021-10-03
fix: correct heroku

idivyanshbansal:
I edited the Heroku.md as same as it looks on the website as you have raised this issue #149


idivyanshbansal:
@vihar sir that issue was asked from the update of Heroku.md and I do the same. I do nothing such as to be considered spam.

PR #269, opened 2021-10-13
feat: Spellcheck property for the input widget

wmdev0808:
-- Add a new entry for property, Spellcheck


Pranay105:
Adding an updated document via gitbook

PR #282, opened 2021-10-18
docs: Fix grammar in application-layout.md

donno2048:
1. "less than minimum" replaced with "less than the minimum"
2. "point of time" replaced with "point in time"


donno2048:
#283

cokoghenun:
Please also work on the feedback given in #283

donno2048:
@cokoghenun done

cokoghenun:
Thanks

PR #300, opened 2021-11-23
docs: geolocation api 

hetunandu:
PR: appsmithorg/appsmith#9295

### DIFF
--- a/framework-reference/appsmith.md
+++ b/framework-reference/appsmith.md
@@ -17,6 +17,7 @@ The appsmith context object contains the following properties
    store: object,
    URL: object,
    user: object,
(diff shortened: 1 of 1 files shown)


framework-reference/appsmith.md
```diff
@@ -63,6 +64,56 @@ This object contains the data of the currently authenticated user.
 }
 ```
 
+### Geolocation
+
+This object contains functions to request the current user location and also the coordinated recived from this request
```
Nikhil-Nandagopal, line 69: I think there are other places this needs to be references like here
https://docs.appsmith.com/core-concepts/writing-code/appsmith-framework

PR #342, opened 2022-02-18
fix: Update aws-ami.md

maxsynnott:
Docs incorrectly state HTTP when it should say SSH


sharat87:
Thank you for your contribution!

PR #382, opened 2022-04-29
Fix: Removed watchtower from docker compose

sum35h:
Removed watchtower from docker-compose (1).yml


sharat87 (commented):
Can we comment the service instead of removing it? That way, if someone _wants_ to enable auto-update, then can just uncomment these lines. Which is what we can instruct people to do, in our docs.

PR #401, opened 2022-05-24
chore: Add redirect urls for the newly created git docs url

AnaghHegde:


.gitbook.yaml
```diff
@@ -9,4 +9,5 @@ redirects:
   setup/instance-configuration/github-login: setup/instance-configuration/single-sign-on-sso/github-login.md
   tutorials/building-a-store-catalog-manager/part-2-using-forms: tutorials/building-a-store-catalog-manager/using-forms.md
   tutorials/building-a-store-catalog-manager/part-3-creating-interactive-views: tutorials/building-a-store-catalog-manager/creating-interactive-views.md
+  core-concepts/version-control-with-git: core-concepts/git-sync.md
```
Pranay105, line 12: It has to be in the following format:
Old_URL : New_filepath.

Opposite of what you've added @AnaghHegde

AnaghHegde: My bad, let me updated the url.

PR #426, opened 2022-07-07
Update embed-appsmith-into-existing-application.md

rlnorthcutt:
Code examples were missing a closing bracket on the meta tag, and the title in the first example was missing in the later ones.


danciaclara:
@harshilp24 can you quickly update these changes on the new version so that we can close this PR. Thanks

danciaclara:
Is being addressed in #806

PR #433, opened 2022-08-05
fix: Update instructions for existing deployments on Heroku

ribtoks:
The problem is that `Dockerfile` in the repository is only a wrapper over `appsmith/appsmith-ce` docker image. If the local image is outdated, docker will not by default pull a fresh one, thus current instructions will actually not lead to the AppSmith instance on Heroku being updated. If you pull image manually before executing these steps, it works fine.

### DIFF
--- a/getting-started/setup/installation-guides/heroku.md
+++ b/getting-started/setup/installation-guides/heroku.md
@@ -100,6 +100,11 @@ Then follow the steps below:
       git clone --branch master https://github.com/appsmithorg/appsmith
       cd ./appsmith/deploy/heroku
     ```
+*   For existing repository, pull the fresh Appsmith image:
+
+    ```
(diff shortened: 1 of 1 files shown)


sharat87 (approved):

sharat87:
Hey @ribtoks, can you accept the CLA from the above comment please? We can merge this in right after that's done.

Sorry for all the delay in getting this moving. Thank you for your contribution! 🎉

jnikhila:
@sharat87 - could you please review this PR? We don't have the v1.4 branch any longer. We can manually make this change to the existing doc. Please confirm if this holds true.

sharat87:
@jnikhila, yeah please go ahead.

jnikhila:
Migrated changes to #786

PR #435, opened 2022-08-15
fix: Remove unused chart images

sharat87:
The fact that these two files have the same name, but with different case, causes serious pain when cloning and working with this repo offline, on filesystems that are case-insensitive, like `macOS`.

See the output of `git clone git@github.com:appsmithorg/appsmith-docs.git`:

```
Cloning into 'appsmith-docs'...
remote: Enumerating objects: 11832, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (13/13), done.
remote: Total 11832 (delta 1), reused 9 (delta 1), pack-reused 11817
Receiving objects: 100% (11832/11832), 1.34 GiB | 7.11 MiB/s, done.
Resolving deltas: 100% (7012/7012), done.
Updating files: 100% (1325/1325), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  '.gitbook/assets/Chart.gif'
  '.gitbook/assets/chart.gif'
```

Then, in that fresh clone, if I do a status:

```
$ git status
On branch v1.4
Your branch is up to date with 'origin/v1.4'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   .gitbook/assets/Chart.gif

no changes added to commit (use "git add" and/or "git commit -a")
```

This change is _impossible_ to get rid of. If I do the suggested `restore`, and then run `git status` again, I get the exact same output, but with `chart.gif` (lowercase) instead.


mohanarpit (approved):

mohanarpit:
@sharat87 Apologies for the delay in getting back to you here. Do we want to delete both the `chart.gif` files or only a single one?

sharat87:
@mohanarpit, removing one would fix the error, so before I raised this PR, I tried searching the repo to see which was actually being used. But neither are, so I deleted both.

Issue #448, opened 2022-09-21
[Docs]: Add information on how to configure default value when server side filtering is enabled


dilippitchika:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the problem

When server side filtering is enabled in the select and multi select widget, users are not aware that the default value format needs to change.

### Describe the improvement

The server side filtering section needs to be updated to inform the users that the default value should be specified in the `{label, value}` format. 

### Why do you think this change is needed?

Since users don't have this information they are confused on how to make defaults work with server side data.

PR #449, opened 2022-09-21
448 docs add information on how to configure default value when server side filtering is enabled

dilippitchika:
Solves the issue #448 

Updated the following widgets
- Select widget
- Multi select widget
to inform the users to update the default value format when server side filtering is turned on.


Pranay105:
Added via Gitbook

PR #527, opened 2022-11-01
Verify anchor redirects

jnikhila:
Added redirect rule to verify if anchor redirect work

### DIFF
--- a/website/vercel.json
+++ b/website/vercel.json
@@ -772,6 +772,19 @@
     {
       "source": "/core-concepts/writing-code",
       "destination": "/reference/appsmith-framework"
+    },
+    {
+      "source":"/v/v1.2/core-concepts/apis/taking-inputs-from-widgets",
+      "destination":"/core-concepts/connecting-to-data-sources/authentication/connect-to-apis#passing-dataparameters-to-api-calls"
+    },
+    {
+      "source": "/tutorials/building-a-store-catalog-manager/part-1-creating-a-simple-view",
+      "has": [
+        {
+          "key": "#accessing-query-results-to-a-widget"
+        }
+      ],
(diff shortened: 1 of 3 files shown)


website/vercel.json
```diff
@@ -772,6 +772,19 @@
...
+      "destination":"/core-concepts/connecting-to-data-sources/authentication/connect-to-apis#passing-dataparameters-to-api-calls"
+    },
+    {
+      "source": "/tutorials/building-a-store-catalog-manager/part-1-creating-a-simple-view",
+      "has": [
+        {
+          "value": "#accessing-query-results-to-a-widget"
+        }
+      ],
+      "destination": "/learning-and-resources/tutorials/building-a-store-catalog-manager/building-a-simple-ui#accessing-query-results-to-a-widget"
```
sharat87, line 783: Redirects cannot handle anchors. The anchor part in a URL is _never_ sent to a server, so the server can't see it's value at all.

Besides, the `has` objects need a `type`, and has to be from (header, cookie, host or query), from https://vercel.com/docs/project-configuration#project-configuration/redirects, so, this likely won't work.

A normal redirect is all we need here. 🙂

```suggestion
    {
      "source": "/tutorials/building-a-store-catalog-manager/part-1-creating-a-simple-view",
      "destination": "/learning-and-resources/tutorials/building-a-store-catalog-manager/building-a-simple-ui"
```

sharat87 (commented):
I see three files in the diff, but only vercel.json has actual changes. Can you check once please? If this branch is corrupted for some reason, please pull changes in `main` branch first, then make a new branch make a PR off of that branch.

website/vercel.json
jnikhila: Okay. Will raise again?

jnikhila:
Will raise a separate PR for this.

PR #546, opened 2022-11-04
fix: Simplify messaging example and fix markup

sharat87:
There's two changes in this.

1. The example code is wrapped in a backtick-string, inside double-braces, but doesn't use any interpolation. So, it's the same thing as directly writing the markup.

2. The code block and the following paragraph need to be part of the 2nd point. For this, they need to be lined up with `We need to embed...`. See the markup and layout in the screenshots to understand this. I can elaborate further on why this is important if needed. Let me know.

## Before
Screenshot 2022-11-04 at 9 59 59 AM

## After
Screenshot 2022-11-04 at 9 58 33 AM

### DIFF
--- a/website/docs/reference/widgets/iframe.md
+++ b/website/docs/reference/widgets/iframe.md
@@ -51,22 +51,18 @@ You can try out the message property by following the steps below:
 
 2. We need to embed a page that is able to send a message with `postMessage()`. In the iframe widget’s settings, copy and paste the following snippet into its **srcDoc** property:
 
-```html
-{{
-    `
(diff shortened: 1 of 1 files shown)


spark-c:
Very helpful screenshots, this makes sense to me. Do you have input @jnikhila ?

spark-c (approved):

jnikhila:
> Very helpful screenshots, this makes sense to me. Do you have input @jnikhila ?

True.

jnikhila (approved):

PR #553, opened 2022-11-09
feat: Table Widget Add Rows

spark-c:
Added docs for new feature: Adding new rows to Tables. Plan to rehaul the inline-editing page, but just added these docs for now to meet the release of the feature.

---

~~Currently a draft PR. Wanted to check in and see if this is an okay way to add this information, or if I should take another route.~~

For now I've got the new feature information added to the inline-editing page, because they share a lot of features/functionality and the work very similarly. I've put it into its own subsection at the bottom, because it has some unique properties that wouldn't make as much sense mixed into the big properties table of the inline-editing section; there is already quite a lot of information on the page and I think mixing the two topics together would make it pretty hard to read.

To accomplish adding a new major section for the "Add new rows" feature, I've bumped all heading levels on the page down by 1 (except for the title). `##` to `###` and so-on. This actually makes the organization of this page more consistent with the rest of the docs in terms of what content is at what heading level.

That was more info than I intended to dump... But does this seem okay?

[First Draft, notion](https://www.notion.so/appsmith/Table-add-new-row-c3ed0d4c5d1e4ee29077e084df3830b8)
[Current docs (Table)](https://docs.appsmith.com/reference/widgets/table/)
[Current docs (Table > inline-editing)](https://docs.appsmith.com/reference/widgets/table/inline-editing)


jnikhila:
@spark-c - As discussed, this looks good to me. Few minor suggestions:
* 'Adding new rows' at the top. So, the page structure could be:
--Inline Editing
    --- Adding new rows
    --- Editing cells
* [Update Mode](https://docs.appsmith.com/reference/widgets/table/inline-editing#update-mode) looks duplicate and should be placed under [properties](https://docs.appsmith.com/reference/widgets/table/inline-editing#update-mode-1).

jnikhila:
Could you please run it through the linter?

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+| **onDiscard** | Action that gets triggered when user clicks discard button for a cell or a new row.                                                       |
+
+## Adding new rows
+
+In addition to editing individual cells, you may allow users to use the Table UI to submit new rows of data. This works similarly to the Inline Editing features.
+
+### How to add a new row
+
+1. Select your Table widget, and ensure that at least one of its columns has its **Editable** property turned on.
+  - Even in new rows, only **Editable** cells can accept user input.
```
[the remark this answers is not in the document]
danciaclara: @spark-c this should probably be added as a note before mentioning the steps to add a new row

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+2. Next, find the **Allow adding a row** property and turn it on as well.
```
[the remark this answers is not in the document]
danciaclara: Also let's not ask the user to find it. Reword this as **Turn on the "Allow Adding a row" property**
Maybe add a screenshot of the property pane to help them find it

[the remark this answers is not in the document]
danciaclara: i think a video of this entire action before mentioning each step would help a lot

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+3. A new button should appear in the Table's UI called 'Add new row'. Click this button to add a row of new cells to the top of the table.
```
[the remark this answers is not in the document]
danciaclara: Table's UI sounds vague. I think this should be 

"A button named **Add new row** will appear in the Table header. Click this button to add a new row"

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+  - If the Table's **Default Values** property has data added to it, the new cells will be pre-populated with those default values. Otherwise, they will be blank.
```
danciaclara, line 256: Let this be a part of the point itself instead of a subpoint. It looks a bit awkward in the vercel preview

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+4. Once the cells have been filled in with desired values, clicking the "Save" button will close the editing mode and execute the Table's **onSave** event. Or, clicking "Discard" will throw away the new cells and trigger the **onDiscard** event.
```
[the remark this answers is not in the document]
danciaclara: Few other points here. Its the **Save Row** button. Apart from mentioning that the Save Row button executing onSave event, we should in some way explicitly mention that for the new row to be saved back to the datasource, user will have to write code to the onSave event and link to the section below.

Also for Discard, let's avoid the phrase 'throw away the new cells' and instead simply mention 'will remove the row from the table'

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+
+### Capturing and using data from new rows
```
danciaclara, line 259: This title is a bit confusing from the user's POV. Can it be something simple like 'Saving new row' ?

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+
+Saving new rows works much like submitting a form. In order for the data to be sent to update the same datasource populating the table, you must use the Table's **onSave** event to execute a query or function that sends the new data to the right place.
```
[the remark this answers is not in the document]
danciaclara: I agree. minor change to what Nikhila suggested. `Saving new rows works much like submitting a form. Use the **onSave** event under the **Adding a row** section on the property pane to execute a query or function that will add the new row to the underlying datasource.`

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+
+<VideoEmbed host="youtube" videoId="09pOAgK9mhk" title="Using newRow to save new table data" caption="Using newRow to save new table data"/>
```
danciaclara, line 263: Introducing the newRow property binding here suddenly without giving much explanation might confuse users. We'll need to rethink this section. Also wherever we are giving examples let's explicitly mention them as an example.

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+{{myAPI_post.run()}}
+```
+
+Assuming that the POST request is successful, the `myAPI` datasource receives the new information, and the Table should contain the new row the next time it updates.
+
+### Properties
+
+The following properties are related to adding new rows:
+
+| Property              | Type        | Definition               | Code Snippet        |
```
danciaclara, line 296: Do we have this Code Snippet column in Properties of other widgets. I'm not sure we should be adding this. Relevant examples when explaining the property in detail should suffice.

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+
+#### Allow adding a row
+
+Turning this on allows users to click a button on the table to start a new row in the table. Users can only input data for columns that are marked as **Editable** in the table properties.
+
+Use the **onSave** event to update the source of the table's data and reflect the user's changes. For example, the **onSave** event might be set to trigger a query that sends the data from a new table row to the database that supplies the table with its values.
+
+#### Default Values
+
+<VideoEmbed host="youtube" videoId="3KozovP7LNk" title="Table | Add new rows | Default values" caption="Set default values for new Table rows"/>
```
danciaclara, line 314: Throughout this page, i noticed the captions below the videos don't stand out as captions. Let's use the standard as on the other pages

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+
+#### newRow
+
+When a new row is being added to the Table, this variable contains a reference to that new row object being added. When there is not a new row in progress, `newRow` is `undefined`.
+
+This is useful for accessing the user input to send it via query once the user is ready to submit the new data.
+
+#### isNewRow
+
+`isNewRow` can be used only within the **Valid** property. With it, you may set specific validation rules for new table rows that does not neccesarily apply to editing existing cells.
```
danciaclara, line 330: I think this is incorrect. It can be used in all validation related properties of the column. i.e Regex, Valid, Error Message, Required. Can we confirm this with @sbalaji1192 ?

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
+A user may make any changes to existing rows, however they are not allowed to add any new row in AE.
+
+### Events
+
+| Events        | Description                                                                                                            |
+| ------------- | ---------------------------------------------------------------------------------------------------------------------- |
+| **onSave**    | Action that gets triggered when the user clicks the save button for a cell or a new row.                               |
+| **onDiscard** | Action that gets triggered when user clicks discard button for a cell or a new row.                                    |
 
 ## What's next?
```
danciaclara, line 350: The What's next section here seems inappropriate. I know it was there before. Maybe we should revisit this as part of rehaul.  It should be either something like tutorials or articles , common errors, troubleshooting, sample apps all related to this topic on inline editing.

danciaclara (changes requested):

website/docs/reference/widgets/table/inline-editing.md
sbalaji1192: Yes. All validation properties have access to the isNewRow variable.

spark-c:
@sbalaji1192 Hi! Do you know what the new release date is for this feature for adding table rows?

sbalaji1192:
@spark-c We don't have an exact date. But it should be available in the upcoming release `1.8.9`

website/docs/reference/widgets/table/inline-editing.md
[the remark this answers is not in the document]
spark-c: @jnikhila I probably wouldn't prefer "Add a new empty row" only because of a tiny technicality -- If the Table's **Default Value** property has been filled, then the new row will not be empty when it is created.

Or if you like, I could change the bit to the above, and add a little `info` block or something colorful to let users know how to fill the **Default Value** bit if they don't want the new row to be empty!

(Also, Dancia's comment has been addressed)

website/docs/reference/widgets/table/inline-editing.md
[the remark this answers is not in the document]
spark-c: Moved the info about **Editable** column settings into its own :::info block before the steps.

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +235,117 @@ This property contains the details of the row that triggered the action (`onSubm
...
 | ------------- | ---------------------------------------------------------------------------------------------------------------------- |
 | **onSubmit**  | Action that gets triggered when the user moves away from editing a cell by pressing the enter key or clicking outside. |
-| **onSave**    | Action that gets triggered when the user clicks the save button.                                                       |
-| **onDiscard** | Action that gets triggered when user clicks discard button                                                             |
+| **onSave**    | Action that gets triggered when the user clicks the save button for a cell or a new row.                                                       |
+| **onDiscard** | Action that gets triggered when user clicks discard button for a cell or a new row.                                                       |
+
+## Adding new rows
+
+In addition to editing individual cells, you may allow users to use the Table UI to submit new rows of data. This works similarly to the Inline Editing features.
```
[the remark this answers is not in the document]
spark-c: - Addressed Nikhila's comment

- Added ticket for removing the Custom update mode

- Also, once this PR is merged with the correct information (the adding rows feature is due to go live soon in the next release) I'm planning to go through and properly restructure this page. I like the order you've proposed for the page structure, how do you feel @jnikhila ?

website/docs/reference/widgets/table/inline-editing.md
spark-c: I think it should be enough for now until the page can be rehauled, if that's okay?

In the instructions there's a short bit to explain what `newRow` is:

> you can use the Table's `newRow` attribute to reference the new row

And I've added a hyperlink on `newRow` to take the user to the place where it is explained in a little more detail.

Let me know what you think, I don't have any good ideas for improving that bit in its current form :/

website/docs/reference/widgets/table/inline-editing.md
spark-c: I've included it to match our [Widget Doc Template](https://www.notion.so/appsmith/Widget-Doc-template-5d7fa61a2e2a4ee888487ba9b43bf487) that all of the (updated) widgets use (there are still some needing updated).

I agree though, I don't think the "Code Snippet" column really adds much to any of the tables. All of the binding properties are accessed in a very consistent way, with `<widget_name>.<property_name>`, and it feels like the Code Snippet column just takes up extra space on the page and makes the table a little squished.

spark-c: > All of the binding properties are accessed in a very consistent way, with <widget_name>.<property_name>

(Of course, the moment I said that, I realize that on *this very page* there is an example of a property that does not follow this pattern (`isNewRow`) 😂 . But, I still have the same opinion.)

website/docs/reference/widgets/table/inline-editing.md
spark-c: I'm not sure what you mean here, on my end all of the captions on every video I've seen appear as plain text. Which is odd of course -- but they do appear this way consistently for me.

website/docs/reference/widgets/table/inline-editing.md
spark-c: I created a ticket for rehauling this page, which i will pick up after this PR; included this point and some other action items

spark-c: #18170

website/docs/reference/widgets/table/inline-editing.md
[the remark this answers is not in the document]
jnikhila: @spark-c - It's getting a bit confusing when you introduced default value. May be you could rewrite it to something that says `adding a new row or a row added with default values if default values are set`. Adding an info is not required. You could also check if you could link to an existing doc that explains Default Value.

website/docs/reference/widgets/table/inline-editing.md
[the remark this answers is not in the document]
jnikhila: Sounds good @spark-c

website/docs/reference/widgets/table/inline-editing.md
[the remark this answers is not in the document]
spark-c: Sure, I've updated the sentence a bit to hopefully be more straightforward (removed the nested bullet point) and I also added a link to the sections above on the page where the doc describes **Default Values**

website/docs/reference/widgets/table/inline-editing.md
spark-c: Reordered the sections so that the Properties and Events table appear first before these sections

jnikhila (approved):
Looks good to me.

jnikhila:
@sbalaji1192 - Could you please review the changes and update us on when to release the doc? CC: @spark-c

sbalaji1192 (dismissed):

sbalaji1192:
@jnikhila Add new row feature is in prod. But not available for the self-hosted versions yet. Should we release the docs once it's also available for the self hosted instance?

jnikhila:
@sbalaji1192 - We are okay with that. I presume the feature is available on the cloud too. Would not having docs cause any issues for cloud users? CC: @spark-c - Please take it ahead if the docs needs to be released.

sbalaji1192:
@jnikhila Yeah it is in the cloud. I was coming from the point where we release the doc and people couldn't find the feature on the self-hosted instances, it might create confusion. @dilippitchika What do you think?

dilippitchika:
If we know when tagging will be done we can launch it then @jnikhila @spark-c

spark-c:
@danciaclara would you be able to take a quick look at this and see if any requested changes are still pending? merge is blocked for the moment, I think it is requiring another check from you :)

website/docs/reference/widgets/table/inline-editing.md
```diff
@@ -235,13 +236,118 @@ This property contains the details of the row that triggered the action (`onSubm
...
+| **newRow** | Binding | This variable contains a reference to the new row object added by the user. | `Table1.newRow` |
+| **isNewRow** | Validation | When writing code for validation properties, this variable indicates whether the cells being validated are part of a new or an existing row. | `isNewRow` |
+
+The subheadings below describe these properties in more detail:
+
+#### Allow adding a row
+
+Turning this on allows users to click a button on the table to start a new row in the table. Users can only input data for columns that are marked as **Editable** in the table properties.
+
+Use the **onSave** event to update the source of the table's data and reflect the user's changes. For example, the **onSave** event might be set to trigger a query that sends the data from a new table row to the database that supplies the table with its values.
```
danciaclara, line 269: I think instead of "update the source of the table's data", can we write  "update the original datasource table to reflect the user's changes"?

danciaclara (approved):

PR #564, opened 2022-11-15
chore: add script to update the index used by the omnibar

akash-codemonk:
Related to https://github.com/appsmithorg/appsmith/issues/17501

Adds a script to update the index used by the omnibar.

The script fetches all the markdown under docs/ folder and parses, uploads the same. It updates the href of links in the markdown to full paths. The images for the omnibar are fetched from `https://raw.githubusercontent.com/appsmithorg/appsmith-docs/main/website/static`. The title is the h1 text in the markdown. 

Also updates the workflow to run the script.


akash-codemonk:
> I see that we are deleting the `package-lock.json` file. Is this intentional and we don't need it? Can you elaborate please? If not, can we restore it please?

Ah no idea. Just did npm install and it probably did the changes. I will revert that back.

sharat87 (changes requested):
Can you run `npm install` on this branch once? Since we added two dependencies to `package.json`, the `package-lock.json` should include corresponding changes as well. But I don't see any non-whitespace changes in that file here.

Just run `npm install` once, and then commit the `package-lock.json` file as is. Then we'll be able to use `npm ci` instead of `npm i` in the workflow.

akash-codemonk:
> Can you run `npm install` on this branch once? Since we added two dependencies to `package.json`, the `package-lock.json` should include corresponding changes as well. But I don't see any non-whitespace changes in that file here.
> 
> Just run `npm install` once, and then commit the `package-lock.json` file as is. Then we'll be able to use `npm ci` instead of `npm i` in the workflow.

Updated

sharat87 (approved):

PR #588, opened 2022-11-21
feat: Adding total records count

Pranay105:
- Added total records count in the table widget.
- Updated .gitignore with vale.ini file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ website/i18n/**/*
 #!website/i18n/fr/**/*
 test-appsmith-docusaurus/.env
 .vale.ini
+vale.ini
 Linter/ValeLinter.md
 styles/Google/Acronyms.yml
 styles/Google/AMPM.yml
--- a/website/docs/reference/widgets/table/README.md
+++ b/website/docs/reference/widgets/table/README.md
@@ -28,7 +28,7 @@ These properties allow you to edit the table itself. All these properties are pr
 | **Default Selected Row**       | Sets the default selected row of the table. It takes the index no. of the row as an input.                                                                                                              |
 | **Default Row Height**         | Sets the height of the row in the table - Default, short or tall.                                                                                                                                       |
 | **Enable multi-row selection** | Allows multiple rows of a table to be selected. The rows are populated in the selectedRows field.                                                                                                       |
-| **Total Record Count**         | The user has to input the total number of records returned as part of a query or API call, which will be displayed in a table. It is only visible when you choose to enable **server-side pagination**. |
+| **Total Record Count**         | It stores the total number of rows returned by a query or API, which helps in calculating the total number of pages on the table when using server-side pagination. |
 
 Let's understand the widget properties in detail.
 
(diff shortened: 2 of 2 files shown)


website/docs/reference/widgets/table/README.md
```diff
@@ -28,7 +28,7 @@ These properties allow you to edit the table itself. All these properties are pr
 | **Default Selected Row**       | Sets the default selected row of the table. It takes the index no. of the row as an input.                                                                                                              |
 | **Default Row Height**         | Sets the height of the row in the table - Default, short or tall.                                                                                                                                       |
 | **Enable multi-row selection** | Allows multiple rows of a table to be selected. The rows are populated in the selectedRows field.                                                                                                       |
-| **Table Record Count**         | The user has to input the total number of records returned as part of a query or API call, which will be displayed in a table. It is only visible when you choose to enable **server-side pagination**. |
+| **Total Record Count**         | The user has to input the total number of records returned as part of a query or API call, which will be displayed in a table. It is only visible when you choose to enable **server-side pagination**. |
```
danciaclara, line 31: Here in the description I think we should make it clear that the total record count is used to control the next page control when using server side pagination.  so you can rephrase it as 

The user can provide the total number of rows returned by a query or API, which helps in calculating the total number of pages on the table when using server-side pagination.

danciaclara (changes requested):

danciaclara (approved):

.gitignore
```diff
@@ -40,6 +40,7 @@ website/i18n/**/*
 #!website/i18n/fr/**/*
 test-appsmith-docusaurus/.env
 .vale.ini
+vale.ini
```
jnikhila, line 43: vale.ini is not the file it's .vale.ini only. Do we need this change?

website/docs/reference/widgets/table/README.md
```diff
@@ -28,7 +28,7 @@ These properties allow you to edit the table itself. All these properties are pr
 | **Default Selected Row**       | Sets the default selected row of the table. It takes the index no. of the row as an input.                                                                                                              |
 | **Default Row Height**         | Sets the height of the row in the table - Default, short or tall.                                                                                                                                       |
 | **Enable multi-row selection** | Allows multiple rows of a table to be selected. The rows are populated in the selectedRows field.                                                                                                       |
-| **Table Record Count**         | The user has to input the total number of records returned as part of a query or API call, which will be displayed in a table. It is only visible when you choose to enable **server-side pagination**. |
+| **Total Record Count**         | It stores the total number of rows returned by a query or API, which helps in calculating the total number of pages on the table when using server-side pagination. |
```
jnikhila, line 31: Could you hyperlink to the server-side-pagination section below?

website/docs/reference/widgets/table/README.md
```diff
@@ -59,20 +59,42 @@ Where `<query_name>` is the query's name created in Step 2.
...
+    SELECT COUNT(*) FROM <table_name>;
+    ```
+ 3. Click on Run and the output will be displayed the response tab below.
+
+Once the `get_count` query is successfully created, enter the following code to bind the query's output in the total records count:
+```
+#Postgres
+{{get_count.data[0].count}}
+```
+<VideoEmbed host="youtube" videoId="p7mH00xp7Nc" title="Using Total Records Count in th Table Widget" caption="Using Total Records Count in th Table Widget"/>
```
jnikhila, line 94: I think adding a video above the steps will be helpful.

.gitignore
Pranay105: For me and Dancia it was vale.ini , that's why I added it. Linter works fine though

PR #594, opened 2022-11-22
fix: context-object-update

harshilp24:


website/docs/reference/appsmith-framework/context-object.md
```diff
@@ -52,6 +55,95 @@ This object contains all the values associated with the current URL that the use
...
+```js
+//{{appsmith.URL.host}}
+host:"app.appsmith.com:111"
+```
+
+#### hostName
+The hostname property of the URL is a **string** that holds the **domain of the URL**. 
+
+
+:::info
```
jnikhila, line 72: I think this could be a part of hostName itself, and we don't need it as info.

website/docs/reference/appsmith-framework/context-object.md
```diff
@@ -52,6 +55,95 @@ This object contains all the values associated with the current URL that the use
...
+
+Example:
+```js
+//{{appsmith.URL.hostname}}
+hostName:"app.appsmith.com"
+```
+
+
+#### fullPath
+A full-path URL designates an EXACT location (such as a page, app, file, etc.). In addition to the domain and TLD, a full-path URL needs to include the protocol, subdomain (such as "app," "support," etc.), path/destination, and possibly a file extension as well as parameters. A full-path can include:
```
jnikhila, line 84: Instead of TLD, please define the abbreviation - Top Level Domain (TLD), and then in the doc below refer to it as TLD abbreviation.

website/docs/reference/appsmith-framework/context-object.md
```diff
@@ -52,6 +55,95 @@ This object contains all the values associated with the current URL that the use
...
+pathName:"/app/demo-app/page1-6324031aa"
+```
+
+#### port
+The port property of the URL is a string that contains the port number of the URL.
+
+Example:
+```js
+//{{appsmith.URL.port}}
+port:"localhost:3000/"
```
jnikhila, line 114: Would it just hold the value as 3000?

website/docs/reference/appsmith-framework/context-object.md
```diff
@@ -52,6 +55,95 @@ This object contains all the values associated with the current URL that the use
...
+```
+#### protocol
+The protocol property of the URL is a string that represents the protocol scheme of the URL, including the ```:```.
+:::info
```
jnikhila, line 118: This could be added as a > Quote instead of info.

website/docs/reference/appsmith-framework/context-object.md
```diff
@@ -52,6 +55,95 @@ This object contains all the values associated with the current URL that the use
...
+
+Example:
+```js
+//{{appsmith.URL.hash}}
+hash:"#n912xhesgo"
+```
+
+#### queryParams
+Query parameters are a predefined set of parameters that are added to the end of a URL.
+They are URL extensions that help define particular content or actions based on the data being delivered. A ```?``` is added at the end of a URL, then a query parameter is added right after it.
```
jnikhila, line 139: It could continue in the same para. Please remove the line break.

jnikhila: Could we rewrite this to make it a bit simpler?

`They are URL extensions that help define particular content or actions based on the data being delivered. A ```?``` is added at the end of a URL, then a query parameter is added right after it.`

Something along the lines:
A URL has all the query parameters appended at the end with a "?" as a separator. 

And then maybe add an example URL as below that you have given.

jnikhila (changes requested):

website/docs/reference/appsmith-framework/context-object.md
harshilp24: Keeping the full form - Top Level Domain at both places

harshilp24:
created new PR - https://github.com/appsmithorg/appsmith-docs/pull/597

Issue #601, opened 2022-11-23
[Docs]: Provide tip on using storeValue when there are multiple variables that need to store values


danciaclara:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Documentation Link

https://docs.appsmith.com/reference/appsmith-framework/widget-actions/store-value

### Discord/slack/intercom Link

https://theappsmith.slack.com/archives/C02K0SZQ7V3/p1666252020048049

### Describe the problem

Many users face performance issues when they make multiple calls to storeValue function to store values. 

### Describe the improvement

We should provide a tip on the best practice in this situation that they should store the values as an object array in one storeValue function call . 

The snippet users are usually writing something like below.
`storeValue("showSelect", true);
storeValue("showTable", true);
storeValue("showButton", true);
.....`
when it should be used as
`storeValue("flags", {
 showSelect: true,
 showTable: true,
 showButton: true
});`

PR #609, opened 2022-11-25
fix: store-value-#601

harshilp24:


website/docs/reference/appsmith-framework/widget-actions/store-value.md
```diff
@@ -188,6 +188,22 @@ If you want to clear the session store value, you must define three parameters a
 }}
 ```
 
-### What's Next
+
+> Your app may face performance issues if you make multiple calls to the ```storeValue``` function to store values. It is **recommended** to use an array object to store values. For instance:
```
jnikhila, line 192: I think a heading would be good for this section. Right now, it looks like a part of clear session store value.

jnikhila: On second thoughts, @danciaclara - Could this be moved to the troubleshooting guide? Do we know what error message is shown to the user. The store value doc can cross-reference the troubleshooting guide. Your thoughts??

jnikhila (changes requested):

website/docs/reference/appsmith-framework/widget-actions/store-value.md
danciaclara: My initial suggestion to the engineering team was that, but this is a performance issue (no error messages. app just slows down), so sort of a like best practice of using store value in this scenario. So let's add it in the doc itself

@harshilp24 let's not explicitly say that users will face performance issues :) just mention something like 

> If you need to store many values, instead of making multiple calls to the ```storeValue``` function, it is **recommended** to use an object array to store the values. For instance:

website/docs/reference/appsmith-framework/widget-actions/store-value.md
```diff
@@ -188,6 +188,22 @@ If you want to clear the session store value, you must define three parameters a
 }}
 ```
 
-### What's Next
+
+> Your app may face performance issues if you make multiple calls to the ```storeValue``` function to store values. It is **recommended** to use an array object to store values. For instance: 
+If you are using something like:
```
danciaclara, line 193: remove this line

website/docs/reference/appsmith-framework/widget-actions/store-value.md
```diff
@@ -188,6 +188,22 @@ If you want to clear the session store value, you must define three parameters a
...
+```js
+storeValue("showSelect", true); storeValue("showTable", true); storeValue("showButton", true); .....
+``` 
+It should be used as:
```
danciaclara, line 197: All values should be assigned in a single call to the storeValue() function.

website/docs/reference/appsmith-framework/widget-actions/store-value.md
danciaclara: The heading can be **Storing multiple values**

danciaclara (changes requested):

website/docs/reference/appsmith-framework/widget-actions/store-value.md
```diff
@@ -99,18 +99,18 @@ In Appsmith, two types of scopes are available for the storeValue function. The
...
 You can set a persistent value by simply calling the store value with two parameters:
 
 * Key - Name of the Key
 * Value - The value for the given key
 
-Let's see how it works:
+Lets see how it works:
 
-* Drag and drop the _**input widget**_ to the canvas.
+* Drag the _**input widget**_ to the canvas.
```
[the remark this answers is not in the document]
harshilp24: added drag and drop in the accept.txt file

website/docs/reference/appsmith-framework/widget-actions/store-value.md
harshilp24: Should I include this inside the tip (green disclaimer) or add best practice title @danciaclara @jnikhila

danciaclara: @harshilp24 no, this is not a callout, just keep it a subheading "Storing multiple values"

harshilp24: done

danciaclara (approved):

jnikhila (approved):

PR #613, opened 2022-11-25
chore: New style for tabs

danciaclara:
Had to also make linter changes to the Sample apps file as that is the only page where we use tabs. And the tab item headings were all in uppercase, changed that and some linter related things

### DIFF
--- a/website/docs/learning-and-resources/sample-apps.md
+++ b/website/docs/learning-and-resources/sample-apps.md
@@ -435,7 +435,7 @@ import TabItem from '@theme/TabItem';
 <div class="containerGridSampleApp">
     <div class="containerColumnSampleApp columnGrid column-one">
         <div class="containerCol">
-            <strong>Get Base64 String of Image</strong>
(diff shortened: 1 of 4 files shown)


website/docs/learning-and-resources/sample-apps.md
```diff
@@ -582,7 +582,7 @@ import TabItem from '@theme/TabItem';
     </div>
 </div>
 
-### UI Components
+### UI components
```
jnikhila, line 585: Please add this as an exception to Microsoft>> HeadingAcronyms. Let me know if you need help.

jnikhila (changes requested):

website/docs/learning-and-resources/sample-apps.md
danciaclara: added

Pranay105 (approved):

jnikhila (approved):

Issue #630, opened 2022-12-01
[Docs]: Connecting to Datasource - Rehaul 


harshilp24:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Documentation Link

https://docs.appsmith.com/core-concepts/connecting-to-data-sources/
https://docs.appsmith.com/core-concepts/connecting-to-data-sources/connecting-to-databases
https://docs.appsmith.com/core-concepts/connecting-to-data-sources/authentication

### Discord/slack/intercom Link

NA

### Describe the problem

Rehaul Ticket

### Describe the improvement

Update the doc

PR #632, opened 2022-12-02
fix: Outline for Connecting to Datasource - Rehaul #630

harshilp24:


website/docs/core-concepts/connecting-to-data-sources/README.md
```diff
@@ -1,5 +1,41 @@
 # Connecting to Datasource
 
+## **Outline**
+
+**Introduction** 
+
+**Prerequisites​**
```
danciaclara, line 7: I think this section is unnecessary unless there is some pre-requisite to connecting to a data source.  Creating an account on Appsmith is a known thing and does not need to be mentioned here

website/docs/core-concepts/connecting-to-data-sources/README.md
```diff
@@ -1,5 +1,41 @@
...
+* Sample API​ (Explain how to use sample API)
+
+**Databases**
+* Information about databases
+* List of supported Database.
+* Sample DB​ (Explain how to use sample DB, both movies and users)
+
+**Security**
+
+**Internal database​(LocalDB)**
```
danciaclara, line 27: I think this should come above Security and lets rename it as 'Connect to local datasource'

website/docs/core-concepts/connecting-to-data-sources/README.md
```diff
@@ -1,5 +1,41 @@
...
+
+**Introduction** 
+
+**Prerequisites​**
+
+Before you begin, you'll need to create an account on Appsmith Cloud, or you can self-host Appsmith locally or on a hosting platform of your choice.
+
+**How to connect to a datasource**
+
+Explain how to connect to a datasource
```
danciaclara, line 13: @harshilp24 could you also give a brief about what is the plan for this section.

danciaclara (changes requested):

danciaclara:
Somewhere on this page let's also link to the Datasource reference page

website/docs/core-concepts/connecting-to-data-sources/README.md
harshilp24: note.

website/docs/core-concepts/connecting-to-data-sources/README.md
harshilp24: updated

jnikhila: @danciaclara - Why we want to specify it to local datasource. User's could use a hosted service/free services too. Local word would be misleading here. I think we should keep is `create a datasource connection`.

website/docs/core-concepts/connecting-to-data-sources/README.md
jnikhila: @harshilp24 - This would be a part of the API section, right? This section is a part of the datasource, which would be about databases, as you have segregated the content between two heads - API and Datasource.

website/docs/core-concepts/connecting-to-data-sources/README.md
```diff
@@ -1,5 +1,41 @@
...
+* Information about APIs
+* List of supported APIs. 
+* Sample API​ (Explain how to use sample API)
+
+**Databases**
+* Information about databases
+* List of supported Database.
+* Sample DB​ (Explain how to use sample DB, both movies and users)
+
+**Security**
```
jnikhila, line 25: This section looks a bit offbeat to me. Because we are not talking about the datasource security setups here. It needn't be a section - just a callout or highlight to read more on security.

website/docs/core-concepts/connecting-to-data-sources/README.md
harshilp24: Something like this @jnikhila @danciaclara 
Screenshot 2022-12-05 at 6 59 01 PM

jnikhila (changes requested):

website/docs/core-concepts/connecting-to-data-sources/README.md
danciaclara: Wait this section is about local db/api right? Thats why I suggested changing the heading. 
https://appsmith-docs-git-connecting-to-datasource630-get-appsmith.vercel.app/core-concepts/connecting-to-data-sources/#connect-to-a-localhost-database-api

On another note, I feel this is redundant info from this page -  https://docs.appsmith.com/advanced-concepts/more/how-to-work-with-local-apis-on-appsmith Can't we have this linked there

website/docs/core-concepts/connecting-to-data-sources/README.md
danciaclara: I agree. we dont need to mention it at all. We have a Security page for this, if users want to check it out

website/docs/core-concepts/connecting-to-data-sources/README.md
danciaclara: ok, but we already have this info under Database and API separately, too, right? and that will make it seem redundant. Let this landing page be only about information that is common to both. things we want our users to know in general about datasources,  how they are global to that org, etc...  this page will primarily have the cards to those subpages and other shared information. 

I think for this ticket, we should also take into account the subpages - Databases and REST APIs and the data there.

harshilp24: noted! will remove the how to connect section, also @danciaclara can you check this preview - https://appsmith-docs-gl03cdez4-get-appsmith.vercel.app/core-concepts/connecting-to-data-sources/,

harshilp24:
Merge conflict - created new PR - https://github.com/appsmithorg/appsmith-docs/pull/691

Issue #668, opened 2022-12-12
[Docs]: Support OnFocus and OnBlur for a few widgets


dilippitchika:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Engineering Ticket Link

https://github.com/appsmithorg/appsmith/issues/6452

### Release Date

v1.8.13

### First Draft

#669 

### Loom video

[Loom](https://www.loom.com/share/4a7eb406ccd74dfba09e56045068ccd7)

### PRD

https://appsmith.notion.site/Widget-onFocus-and-onBlur-events-36fc98f9d0d04ce0b369247e91e01e5b

jnikhila:
Opened as we'll merge the PR for changes once the engineering ticket is released. @dilippitchika - FYI

jnikhila:
Opened as we'll merge the PR for changes once the engineering ticket is released. @dilippitchika - FYI

PR #671, opened 2022-12-13
fix: 668 widget focus blur

jnikhila:
Raised a new PR to add changes related to #668 - onFocus and onBlur events.


dilippitchika (approved):

dilippitchika:
@jnikhila yes, in the issue i am making sure i mention the version in which any feature will go live since i don't have a control on the release date. 

I think none of the devs are aware of the release date directly but only the version. We should consider changing the issue template to have that detail and based on version you can auto-merge all approved issues after go live for that version.

jnikhila:
> @jnikhila yes, in the issue i am making sure i mention the version in which any feature will go live since i don't have a control on the release date.
> 
> I think none of the devs are aware of the release date directly but only the version. We should consider changing the issue template to have that detail and based on version you can auto-merge all approved issues after go live for that version.

Sure. @dilippitchika. I'll do the changes to the issue templates.

PR #673, opened 2022-12-13
feat: Add setTimeout docs

spark-c:
This PR adds a `setTimeout` section to the existing [**Intervals (Time Events)**](https://docs.appsmith.com/reference/appsmith-framework/widget-actions/intervals-time-events) page, and also reformats the existing sections to match some of our newer framework functions docs pages.

Resolves [appsmithorg/appsmith#18163](https://github.com/appsmithorg/appsmith/issues/18163)


website/docs/reference/appsmith-framework/widget-actions/intervals-time-events.md
```diff
@@ -2,35 +2,82 @@
...
+Interval events (timing functions in JavaScript) allow the user to periodically run API and DB queries. You can configure these by using the `setInterval` and `clearInterval` functions.
 
 <VideoEmbed host="youtube" videoId="ByE3aqlQ1pE" title="How To Use The SetInterval & ClearInterval Functions To Auto-Update Widgets" caption="How To Use The SetInterval & ClearInterval Functions To Auto-Update Widgets"/>
 
-
-
 ## setInterval()
 
-The setInterval() method executes a code snippet with a fixed time interval between the calls, i.e, it executes a trigger callback at a given interval. Here’s how you can use it while dynamically binding queries onto widgets:
+`setInterval()` executes a callback function with a fixed time interval between the calls. This function is analogous to the [JavaScript `setInterval` function](https://developer.mozilla.org/en-US/docs/Web/API/setInterval).
```
danciaclara, line 13: do we have to mention that it is the same as the browser's function and link to the docs there?  Even if we are, let's not say JavaScript function, instead 'Window setInterval() method' 

Same for the other two functions in this page

danciaclara (changes requested):
This structure looks a bit odd, can we reduce the heading style of Signature, Arguments, and Example so that they don't show up on the right side multiple times. 

Screenshot 2022-12-13 at 23 07 10

website/docs/reference/appsmith-framework/widget-actions/intervals-time-events.md
spark-c: rm'd the bits about the JS functions since a couple of them do have some differences to how they work -- otherwise should be good to go :)

danciaclara (dismissed):

danciaclara (approved):

PR #676, opened 2022-12-14
fix: decrease size of headers globally

tjtln:


jnikhila:
Thanks @tjtln for your contribution. We have planned style changes lined up in our roadmap and will coming these changes as part of that.

PR #684, opened 2022-12-15
fix:sql-modes

harshilp24:


website/docs/reference/datasources/querying-mysql.md
```diff
@@ -53,15 +53,39 @@ MySQL databases can be queried using the standard [SQL syntax](https://dev.mysql
...
-Follow the guide on [how to use prepared statements](/learning-and-resources/how-to-guides/how-to-use-prepared-statements.md) for efficient and secured data transactions.
+```SQL_MODE``` is a system variable in MySQL that controls the behavior of the MySQL server. It can be used to configure the server to be strict or forgiving when accepting input data, enable or disable standard SQL conformance, or provide better compatibility with other databases. [Strict mode](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sql-mode-strict) can help you ensure the integrity of your data by preventing the database from automatically inserting default values for missing or invalid data. 
 
-## Using Queries in Applications
+#### Setting SQL modes
+
+To change the SQL mode at runtime, you can use a **SET** statement to set the `GLOBAL` or `SESSION` ```sql_mode``` system variable:
+
+```js
+SET GLOBAL sql_mode = 'mode1,mode2,...';
```
subrata71, line 67: Switching to the strict mode in this way will require the knowledge of each mode from the user. But there's an easy way to switch to strict mode i.e. `SET sql_mode = TRADITIONAL`. 
Don't you think it comes in handy?

harshilp24: @jnikhila @danciaclara should we keep this as a general explanation or add info for a specific mode?

danciaclara: If the recommended SQL mode is strict mode then we should mention this.

harshilp24: for reference: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-important

jnikhila: @harshilp24 - We can add that it's recommended to use strict mode and then link to the official documentation of Mysql?

jnikhila (approved):

website/docs/reference/datasources/querying-mysql.md
```diff
@@ -53,15 +53,39 @@ MySQL databases can be queried using the standard [SQL syntax](https://dev.mysql
...
+SET SESSION sql_mode = 'mode1,mode2,...';
+```
+where **mode1**, **mode2**, etc. are the specific SQL modes that you want to enable or disable. To switch to strict mode, it is recommended to use the TRADITIONAL mode by setting ```SET sql_mode to TRADITIONAL```.
```
subrata71, line 70: The query is not correct. 
It should be ```SET sql_mode = TRADITIONAL```

harshilp24: my bad :), updated

danciaclara (approved):

subrata71 (dismissed):
LGTM

jnikhila (approved):

danciaclara (dismissed):

subrata71 (approved):

jnikhila (approved):

Issue #311, opened 2021-12-22
Update phrase in introduction section


nayan-rafiq:
In the introduction page, we've the following

> Here are five simple steps to create an app in Appsmith:

By steps, usually users understand a sequence of actions that someone need to complete in order to achieve something. The described points are not steps but are functionalities/features/use cases. Changing the text to something else seems more appropriate in this case.

nayan-rafiq:
It can be like the following:

How does Appsmith work? -> What you can do with Appsmith?

Here are five simple steps to create an app in Appsmith: -> Here are top five use cases you can do with Appsmith:

kasarchou:
Hello @nayan-rafiq i am trying to find the phrases you are talking about but it seems like the have already been resolved?

AS-Laguna:
I'm available to make these changes.

danciaclara:
was updated by https://github.com/appsmithorg/appsmith-docs/pull/701

PR #686, opened 2022-12-16
fix: Update intro phrases 

carbonara-mac:
Updated intro as stated in issue #311


danciaclara:
@carbonara-mac Thank you for catching that and also putting in this PR :)  What do you think about using 'Why Appsmith?' to have a shorter heading? Would that make sense as if we are trying to convey " why appsmith because you can connect datasource, build ui, .... build crud apps, etc" ?

danciaclara (approved):

danciaclara (approved):

jnikhila:
Changes are added as part of #701

PR #723, opened 2023-01-05
fix: Table Widget Rehaul

spark-c:
Rehaul of Table widget docs, including:

- table/README.md
- table/column-settings.md
- [/core-concepts/data-access-and-binding/displaying-data-read/display-data-tables](https://docs.appsmith.com/core-concepts/data-access-and-binding/displaying-data-read/display-data-tables) (removed, content moved to table/README)

Still making small fixes here-and-there as well.


danciaclara:
I think this should have been here before. Apologies for the inconvenience. Will ensure it is done for every rehaul ticket. Please rearrange the content based on the structure below.

# Table
   [Intro]

## Properties
     [Properties table] 

## Events
     [Events table]

## Display data in Table
      [Intro]
      [Explain relevant properties]
**Table Data**
**Columns**
**Add new columns**

## Server side pagination
      [Intro and Explanation, Add youtube video if available]
**pageNo**
**pageSize**
**Total Records**
### Offset based pagination
### Key based pagination

## Server side searching
      [Intro]
**searchText**
###
        [Explain concept with an example]

## Server side filtering
      [Intro]
      [Explain concept with an example]

## Work with Table data
      [Intro]
**selectedRow**
**selectedRows**
**triggeredRow**
**selectedRowIndex**
**selectedRowIndices**
**filteredTableData**

### Transform table data
### Update table data in real time

## Table header options
      [Intro]
      [Mention relevant properties]
**Search**
**Filters**
**Download**
      [Mention CSV separator]
**Add a row**
     [Link to adding rows section in inline editing]
**Pagination**

website/docs/reference/widgets/table/README.md
```diff
@@ -72,171 +83,144 @@ Controls how overflowing contents of the column are handled. When turned on, the
...
-| **filteredTableData**  | Contains the data of the rows left after applying any selected filters, sort rule, or search terms.                  | `{{<table_name>.filteredTableData}}`   |
-| **pageNo**             | Contains the current page number that the user is on. APIs can use it for pagination                                         | `{{<table_name>.pageNo}}`              |
-| **pageSize**           | Contains the number of rows that can fit inside a page of the table. Changes along with the height & row height of the table | `{{<table_name>.pageSize}}`            |
-| **searchText**         | Contains the search text entered by the user in the Table                                                                    | `{{<table_name>.searchText}}`          |
+#### Client side search
 
-For the following examples of binding properties, consider a table called `Table_1` with the table data given below:
+Client side search controls whether the user can search the entire data set for their query, or only what's currently shown in the table widget.
```
danciaclara, line 136: If we are adding this ensure you explain and give an example more than what is already there in the Properties table above.  If it will be added, let it right above Server side pagination in the structure shared

website/docs/reference/widgets/table/README.md
```diff
@@ -72,171 +83,144 @@ Controls how overflowing contents of the column are handled. When turned on, the
...
+#### Column sorting
 
-And this is how the data looks in the table widget:
+Toggles whether table columns are sort-able. When turned on, users may click column headers to sort the table rows by that column's value. This setting only applies while the app is in View mode (where the app is deployed).
 
-![](/img/table\_sample\_data.png)
+<VideoEmbed host="youtube" videoId="hmi7BaF3jFI" title="Table | Column Sorting" caption="Use this setting to enable sorting rows by column value."/>
 
-**selectedRow**
+#### Visible
```
danciaclara, line 144: i think we don't need to explain this in detail. Whatever is mentioned in the Properties table is enough

website/docs/reference/widgets/table/README.md
```diff
@@ -72,171 +83,144 @@ Controls how overflowing contents of the column are handled. When turned on, the
...
+When you tick the checkbox, `Visible` is set to `true`, and the table becomes visible in the app.
 
-Drag a text widget onto the canvas and bind it to a selected table row of `Table_1`.\
-Open the property pane of the text widget and add the following snippet to its label:
+<VideoEmbed host="youtube" videoId="Jb5bNVhFoRE" title="Visible" caption="Visible"/>
 
-```
-{{Table_1.selectedRow}}
-```
+#### CSV separator
```
danciaclara, line 159: again doesn't need so much explanation or a video.  Just make a mention in the Download part under Table header options

website/docs/reference/widgets/table/README.md
```diff
@@ -245,152 +229,211 @@ The following video shows how to bind a text widget to `Table_1` using seledRowI
...
 ```
 
-Where `<table_name>` is the name of your table.
-
 The following video shows how to bind a text widget to `Table_1` using searchText.
 
 <VideoEmbed host="youtube" videoId="vn6zx7zMeUs" title="searchText" caption="searchText"/>
 
-### Events
+### Header options
```
danciaclara, line 295:
```suggestion
### Table header options
```

danciaclara (changes requested):

website/docs/reference/widgets/table/README.md
spark-c: I removed it for now -- I'd like to explain it better since it's not clear what it does, however I'll need to follow up with someone and update in a subsequent PR at some point I think. I spent a good while experimenting around with this setting weeks ago and was never able to get it to do anything meaningful.

spark-c:
@danciaclara Added `pageOffset` property to the table, added a subheading for it, and updated the instructions for Server Side Pagination

website/docs/core-concepts/data-access-and-binding/displaying-data-read/display-data-tables.md
```diff
@@ -1,98 +0,0 @@
-# Display Data in Tables
```
danciaclara, line 1: Just adding it here so we don't forget to add a redirect from this page to the new section. Also in the Dusplay data page, you can cross-reference to this section so that people can find it from there

website/docs/reference/widgets/table/README.md
```diff
@@ -1,36 +1,90 @@
...
+# Table
 
+The table widget displays data in rows and columns. You can display data from an API in a table, trigger an action when a user selects a row, and even work with sizable paginated data sets. Once a table is connected to a datasource, you can also use [inline-editing](/reference/widgets/table/inline-editing) features to make changes and additions to the table data directly in the widget.
 
-:::info
-If you already know the basics and want to learn how to display data and handle pagination inside a table, [Read this guide](/core-concepts/data-access-and-binding/displaying-data-read/display-data-tables#pagination).
-:::
+<VideoEmbed host="youtube" videoId="-rzePEV2QQ8" title="How to use Table Widget" caption="How to use Table Widget"/>
 
 ## Properties
```
danciaclara, line 12: Please break down the table into multiple tables if you feel the information is too much for one table. But make sure you don't spread out the property table for eg: Styles. You can refer to the platform to see how the properties are being segregated

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
+By clicking the "+ ADD A NEW COLUMN" button, you can insert your own custom column into your dataset. These can be used for a wide variety of purposes, including adding button- or checkbox-type columns alongside your data. You can adjust the settings for this column just like any other.
 
-#### Inline editing
+## Server side pagination
 
-Inline editing allows users to edit cell contents in the table columns. You can enable it for the whole column, or you can use code to enable it for only particular cells. You can also allow users to add new rows of data to the Table. Read more about inline editing [here](./#inline-editing).
+Tables are often required to display large data sets from [queries](/core-concepts/data-access-and-binding/querying-a-database) and [APIs](/core-concepts/connecting-to-data-sources/authentication), but browsers can't always load all the data present in the database. To paginate this data and request smaller chunks of data at a time:
```
danciaclara, line 133: here find out the error that users see when they try to display large amount of data in a table and cross reference it in the troubleshooting guide.

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
+`pageNo` gets the page number of the table that the user is currently viewing. This property can be used by APIs for pagination. To use this property in a widget, enter the code snippet given below:
 
-<VideoEmbed host="youtube" videoId="p7mH00xp7Nc" title="Using Total Records Count in th Table Widget" caption="Using Total Records Count in th Table Widget"/>
+```javascript
+{{<table_name>.pageNo}}
+```
+
+The following video shows how to bind a text widget to `Table_1` using `pageNo`.
+
+<VideoEmbed host="youtube" videoId="DqKok2cCJk0" title="pageNo" caption="pageNo"/>
```
danciaclara, line 149: don't think we require a video for this.

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
+Tables are often required to display large data sets from [queries](/core-concepts/data-access-and-binding/querying-a-database) and [APIs](/core-concepts/connecting-to-data-sources/authentication), but browsers can't always load all the data present in the database. To paginate this data and request smaller chunks of data at a time:
 
-#### Total record count
+1. Enable the server-side pagination property in the table
+2. Call the API / query via the **onPageChange** event
+3. Configure pagination in the API / query using the [Offset](#offset-based-pagination) or [Key-based](#key-based-pagination) pagination method.
 
-Total record count stores the total number of rows in the table. It's useful in pagination as it helps in determining the number of pages, thus enabling/disabling the next/previous page control in the table. It's only visible when you enable **server-side pagination**.
+#### pageNo
```
danciaclara, line 139: before diving into the properties that are used for achieving server-side pagination, mention that

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
+To bind a widget using this property, enter the code snippet given below:
+
+```javascript
+{{<table_name>.pageSize}}
+```
 
-To get the Total record count of your data, follow the steps below:
+The following video shows how to bind a text widget to `Table_1` using pageSize.
+
+<VideoEmbed host="youtube" videoId="XkFJQh4vcCw" title="pageSize" caption="pageSize"/>
```
danciaclara, line 174: again I think a video is uncessary here

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
+
+#### Total records
+
+The total records field stores the total number of rows in the table. It's useful in pagination as it helps in determining the number of pages, and then when to turn on/off the page control buttons in the table header. It's only visible when you enable **Server Side Pagination**.
+
+<VideoEmbed host="youtube" videoId="p7mH00xp7Nc" title="Using Total Records in the Table Widget" caption="Using Total Records in the Table Widget"/>
+
+To get the total record count of your data, follow the steps below:
```
danciaclara, line 182: condense this information. Usually we notice users don't have the patience to read through all of this info to understand what they need to do to achieve something. The video is already there to explain things in detail. so for textual content instead of explaining every step in detail,  you can assume that the query already exists and then what the user should bind and where

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
+```
 
+As an API request with URL parameters:
 ```
-{{Table_1.selectedRow}}
+https://mock-api.appsmith.com/users?name={{Table1.searchText}}
 ```
 
-<VideoEmbed host="youtube" videoId="0Pl7p1sA1fY" title="selectedRow" caption="selectedRow"/>
+### Server side filtering
```
danciaclara, line 259: this does not have to be under searching

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
 
-![](/img/table\_sample\_data.png)
+This method uses a value in the response of the API as the key to the next API call. This can be configured in the API settings by providing the Next & Previous URLs that the API should execute **onPageChange**.
 
-**selectedRow**
+![](</img/pagination_(2)_(2).gif>)
 
-A table you have created might contain a large amount of data. But when you want to get the data of a single row of the table, the selectedRow property comes into play.\
-selectedRow is an internal property that fetches the data of the table row that the user selected. You can bind different widgets to each table row using this property.
+## Server side searching
```
danciaclara, line 222: the content in this section appears a bit verbose. need to condense

website/docs/reference/widgets/table/README.md
```diff
@@ -60,183 +114,231 @@ Now, the table widget should be populated with the data coming from the query.
...
-{{<table_name>.selectedRow.<column_name>}}
+https://mock-api.appsmith.com/users?gender={{genderDropdown.selectedOptionValue}}
 ```
 
-This gives you the data for `<columnname>` of the selected row. Using this snippet in `Table_1`, suppose you want to bind the task of the selected row to the text widget. To do this open the property pane of the text widget and add the following to its label:
+## Work with table data
+
+These properties and concepts are used to access your table data with code and process the dataset so the values are correctly formatted in the table widget.
+
+#### selectedRow
```
danciaclara, line 283: a general suggestion when explaining properties - give heading with some context around using the property. for eg **Accessing the selected row**  and directly dive into examples. Only provide videos where you feel users will not understand how to use a feature.

danciaclara (changes requested):

website/docs/reference/widgets/table/README.md
spark-c: Do you mean it should be at the same heading level as **Server side searching**? I updated them to be equal levels, let me know if you mean something different.

website/docs/reference/widgets/table/README.md
spark-c: I think the single table may be okay, since the properties appear in the same order as in the properties pane and users can use the browser Find function to help; breaking it into smaller tables might just make it more confusing. I'm open to alternative ways to try though, just let me know @danciaclara

website/docs/reference/widgets/table/README.md
spark-c: Do you mean the bit about **Total Record Count** only being visible when **Server side pagination** is enabled? Added, let me know if you mean something different.

website/docs/reference/widgets/table/README.md
spark-c: I agree in general, however in this particular case I think it might be best to keep these steps for **Total records**.

#### info-dump, sorry!

For most properties, a user will already have a query set up and the beginning steps aren't necessary, as you said. However for this one, I think it will be very rare for a user to have _already set up_ their own query specifically to count the number of records. And then, while basic `SELECT` queries are fairly common-knowledge, I think it will be more rare for a user to already know the syntax for using `SELECT COUNT(*)`.

---

More generally (kinda off-topic for this review comment), I also wanted to mention some thoughts I have about videos. They're super useful, though I try to use them strictly as supporting material and keep all of the information accessible via text. I'm a little hesitant to assume that users will have watched any given video, because:

- personally, videos are my last resort when using docs because they tend to take much longer to convey the information
- most of our videos aren't narrated and lack explanation for what/why we're doing (it's not usually hard to guess, but still)
- it's more difficult to find/skip in the video to rewatch a certain bit, than it is to read the step again
- we record in 720p and the appsmith code font is thin, making it kinda hard to read text
- users with spotty or unstable internet may have extra difficulty
- it's probably mostly safe to assume that our users wouldn't be using the appsmith platform if they are visually-impaired or require screen-readers, but I still try to make sure the text could work on its own if needed.

website/docs/reference/widgets/table/README.md
spark-c: Updated, how's this?

> Server side searching is useful for reducing unnecessary results from queries: rather than requesting lots of data from the server and then filtering it on the client, you can pass search terms to the server so it only fetches relevant results in the first place. This can significantly improve response times when working with large data sets.

danciaclara (approved):

PR #725, opened 2023-01-05
fix: 724 migrate to v5 mongodb

jnikhila:
Added a troubleshooting guide to migrate to v5.o MongoDB


website/docs/help-and-support/troubleshooting-guide/action-errors/mongodb-errors.md
```diff
@@ -0,0 +1,22 @@
+---
```
mohanarpit, line 1: @jnikhila Instead of putting this in the troubleshooting guide, do you think this should be a dedicated upgrade page (like the MongoDB upgrade page)?

jnikhila: Are we planning to add more details to the documentation for Upgrade that will be specific to Appsmith? If that's the case, then we can add it as a dedicated MongoDB Upgrade page.

mohanarpit (approved):

PR #743, opened 2023-01-12
fix: Added Cross-reference to Google Maps

jnikhila:
To ensure that users on self-hosted Appsmith can configure Google Maps to use Maps widget.

### DIFF
--- a/website/docs/reference/widgets/maps.md
+++ b/website/docs/reference/widgets/maps.md
@@ -2,6 +2,10 @@
 
 The Map widget is used to display lat, long data on a google map and allows users to select locations on a map as lat, long.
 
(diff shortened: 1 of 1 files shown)


website/docs/reference/widgets/maps.md
```diff
@@ -2,6 +2,10 @@
 
 The Map widget is used to display lat, long data on a google map and allows users to select locations on a map as lat, long.
 
+:::info
```
danciaclara, line 5: Mark it as Important.

danciaclara (changes requested):

danciaclara (approved):

PR #756, opened 2023-01-17
fix: Redis does not support all commands

harshilp24:
- Removed `"Please refer to the official [documentation](https://redis.io/commands) for the list of commands that are supported." `
- Lintr Fixes

### DIFF
--- a/website/docs/reference/datasources/querying-redis.md
+++ b/website/docs/reference/datasources/querying-redis.md
@@ -7,17 +7,17 @@ sidebar_position: 14
 The following document assumes that you understand the [basics of connecting to databases on Appsmith](/core-concepts/connecting-to-data-sources/connecting-to-databases.md#connecting-to-a-database). If not, please go over them before reading further.
 :::
 
-## Connection Settings
+## Connection settings
 
 Appsmith needs the following information to connect to a Redis instance.
 
 ![Click to expand](/img/redis-datasource-form.png)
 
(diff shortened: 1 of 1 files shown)


website/docs/reference/datasources/querying-redis.md
```diff
@@ -7,17 +7,17 @@ sidebar_position: 14
...
 
 ![Click to expand](/img/redis-datasource-form.png)
 
-After filling up the three fields as described above, click on the "Test" button to verify the configuration and click "Save".
+After filling up the three fields as described preceding, click the "Test" button to verify the configuration and click `Save`.
 
 ## Querying Redis
 
-The Redis query pane can be used to run any of the commands supported by Redis. The response from this command is displayed in the result window. Please refer to the official [documentation](https://redis.io/commands) for the list of commands that are supported. A few examples are listed below.
+The Redis query pane can be used to run any of the commands supported by Redis. The response from this command is displayed in the result window. A few examples are listed below.
```
danciaclara, line 20: Also remove 'A few examples are listed below' . In our docs we are just giving out the commands the users can use. we are not giving any examples as such. When we rehaul this page, we should work on an example for each type of command.

danciaclara (changes requested):

website/docs/reference/datasources/querying-redis.md
harshilp24: done

danciaclara (approved):

PR #779, opened 2023-01-24
fix: Update Actions section

harshilp24:


danciaclara (dismissed):

danciaclara (approved):

danciaclara (approved):

ravikp7:
I found more instances in the documentation that need to be updated. Below are few examples, in these sections old action selector UI and its limitations are mentioned. Please check once and let me know if we need to jump on a call.
1. https://appsmith-docs-doa9grbw2-get-appsmith.vercel.app/core-concepts/writing-code/workflows#complex-workflows
2. https://appsmith-docs-doa9grbw2-get-appsmith.vercel.app/reference/widgets/button#trigger-actions-with-buttons

ravikp7:
One more thing that we need to update, with the redesign we’re encouraging users to use then/catch statements to setup success and error blocks instead of callback arguments style that we’ve been mentioning in the docs.

jnikhila:
@danciaclara - @ravikp7 has pointed out:
>I found more instances in the documentation that need to be updated. Below are few examples, in these sections old action selector UI and it’s limitations are mentioned. Please check once and let me know if we need to jump on a call.
https://appsmith-docs-doa9grbw2-get-appsmith.vercel.app/core-concepts/writing-code/workflows#complex-workflows
https://appsmith-docs-doa9grbw2-get-appsmith.vercel.app/reference/widgets/button#trigger-actions-with-buttons

As part of Rehaul we are anyways only adding screenshots when absolutely necessary. There are below ways in which we can handle this:
* Remove the screenshots from above pages so that users are not confused as suggested by Ravi and work on the pages as part of Rehaul pipeline.
* Update the screenshots
* Leave them as is 

Your thoughts?
CC: @harshilp24

PR #740, opened 2023-01-11
fix: 660 rehaul writing code

jnikhila:
- Landing page changes for writing code
    - Restructured and rewrote the landing page
    - Moved relevant content to the landing page for JavaScript Editor from JS Objects page
    - Removed Appsmith Framework from the sidebar (under Writing Code)
- Landing page changes for Appsmith Framework

PR #785, opened 2023-01-25
fix: replaced missing videos

spark-c:
A couple of the videos on the Writing Code page were replaced, along with a few insignificant linter fixes.


jnikhila:
It'll be fixed as part of PR #740

PR #741, opened 2023-01-11
feat:  Parent message

Pranay105:
- Added content for parent message

PR #807, opened 2023-02-01
fix: Added postWindowMessage reference image

spark-c:
Added reference image: https://docs.appsmith.com/reference/appsmith-framework/widget-actions/post-message#types-of-communication


danciaclara (changes requested):
@spark-c Can you update the image based on comment shared here - https://github.com/appsmithorg/appsmith-docs/pull/741#issuecomment-1383935749

website/docs/reference/appsmith-framework/widget-actions/post-message.md
```diff
@@ -114,6 +116,8 @@ When the iframe receives your message, it puts the message text into the `#targe
 
 #### From embedded page to Iframe widget
 
+![](/img/postmessage_child_outgoing.png)
```
danciaclara, line 119: Screenshot 2023-02-09 at 16 44 17
I think this one can be removed for now, because the image shows communication from embedded page to Appsmith which is not possible today. The section talks about communication from element inside the Iframe widget to the the Iframe widget. so we can skip this diagram for now and add it later
```suggestion

```

danciaclara (changes requested):

danciaclara (approved):

PR #808, opened 2023-02-02
fix: Select Widget - Rehaul

harshilp24:


website/docs/reference/widgets/select.md
```diff
@@ -1,106 +1,89 @@
 # Select
 
-‌Select / Dropdown widget is used to capture user input/s from a specified list of permitted inputs.
+‌Select / Dropdown widget is used to capture user input/s from a specified list of permitted inputs. It offers a simple and intuitive interface for users to select a single option from the list, making it an essential tool for capturing user inputs in various forms and applications.
```
danciaclara, line 3: can we combine both statements into one, as both mean the same? Also couple of other things to keep in mind. Let's not explain properties that do not explanation like position, text, etc . For such properties, the description in the describe them well in the properties table.  For the properties that you think that need an explanation, put them under a heading that is meaningful. For eg , how we categorized and explained properties related to displaying data in the select widget. Similarly, if you want to explain about the 'Allow searching' property, you can club it under a sub-heading called **Search and filters** and  put everything related under that. Similarly things like selectedOptionValue should be clubbed under a topic and then explained. Like Accessing selected options or something like that.  Giving property names and headings don't make sense to the user and they will be overwhelmed with the information if there is no proper structuring in the documentation

danciaclara (changes requested):

danciaclara (approved):

PR #809, opened 2023-02-02
fix: GAC | Added documentation for Default Role for All Users

infinitetrooper:
The default role for all users provides powerful control over what a user can see and do in an instance. Admins can use it to provide really fine-grained base level access to their users.


danciaclara:
@infinitetrooper It's not clear where the admin can find this role. Is it under the Others tab?

Pranay105:
@danciaclara I had a conversation with Akhil. This has to be placed right after the introduction as it appears in the roles list, not under any permissions tab.

Pranay105 (approved):

PR #811, opened 2023-02-03
fix: updated Google/GitHub SSO pages to include steps for Admin Settings UI

### LINKED ISSUE #506: [Docs]: Updating Google & Github SSO documentation
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Documentation Link

https://docs.appsmith.com/getting-started/setup/instance-configuration/authentication/google-login

### Discord/slack/intercom Link

https://theappsmith.slack.com/archives/CGBPVEJ5C/p1666768301142219

### Describe the problem

The last step in docs page for Google/ Github SSO requires user to insert the Client ID and Client Secret in the docker image whereas currently the last step is to insert the Client ID and Client Secret in the instance itself. And this doc is linked to home page for setting-up Google/ Github SSO within the instance, and hence can create confusion among users.
 

### Describe the improvement

Should replace the last step in docs page to show that Client ID and Client Secret should be inserted in the instance itself

### Why do you think this change is needed?

This will avoid confusion among users trying to set-up SSO via Google/ Github

spark-c:
Added UI method for providing Client ID and Client Secret for configuring SSO.

Docs pages: [Google](https://docs.appsmith.com/getting-started/setup/instance-configuration/authentication/google-login) | [Github](https://docs.appsmith.com/getting-started/setup/instance-configuration/authentication/github-login)

Resolves #506


website/docs/getting-started/setup/instance-configuration/authentication/github-login.md
```diff
@@ -27,28 +27,43 @@ To enable GitHub Sign in, login to your [GitHub Account](https://github.com)
...
-**6.** [**Restart the Appsmith instance**](../)
+    ```bash
+    # Example configuration in docker.env file
+    # ********* Github OAUth **********
+    APPSMITH_OAUTH2_GITHUB_CLIENT_ID=YOUR_GITHUB_CLIENT_ID
+    APPSMITH_OAUTH2_GITHUB_CLIENT_SECRET=YOUR_GITHUB_CLIENT_SECRET
+    # ******************************
+    ```
 
-:::tip
```
danciaclara, line 52: why are we removing this tip?

danciaclara (changes requested):
@spark-c let's not remove anything in the current document. We only have to add a line where necessary that the the client id and secret can also be updated through Admin Settings in the UI.

spark-c:
No content was removed, only reordered; I broke step 5 into two possible methods (UI vs Docker.env). Since step 6 was just restarting the instance (which has a different method for either UI or docker.env), I combined it into the step 5 part.

website/docs/getting-started/setup/instance-configuration/authentication/github-login.md
spark-c: Oops; I had a reason for turning this tip into plain text (it appears at the bottom of doc) but I don't remember exactly why I did it. Looking at it again now, it should probably stay as a tip and move up a bit, I'll adjust that

danciaclara (approved):

PR #846, opened 2023-02-13
Update docusaurus.config.js

Nikhil-Nandagopal:


danciaclara:
Need more clarity on how we have set up Algolia

PR #854, opened 2023-02-13
custom ssl info- do not set custom domain enc

sum35h:
The doc’s look a bit ambiguous for setting up custom certs and a lot of users end up configuring APPSMITH_CUSTOM_DOMAIN after setting up custom certs.


danciaclara:
@sum35h Just to make sure that I understand this correctly -  So we no longer want the users to use APPSMITH_CUSTOM_DOMAIN for custom certificates, which was used earlier for Let's encrypt? Meaning can users still configure it for Let's encrypt?  in which case we should be explaining this in this doc. in general, I feel the the callout is not clear, especially for new users.

sum35h:
@danciaclara My bad, this was a mistake on my end. It's fine as it is. Cloding the issue.

PR #881, opened 2023-02-18
fix: Copy button copies four spaces in the start

sharat87:
The copy button on this code block is copying the curl command with four additional spaces in the start. This PR fixes that.


danciaclara:
@sharat87 I changed it because completely removing those spaces makes the code block look like below in the preview
Screenshot 2023-02-20 at 10 27 30

I think those spaces are required because it is inside <li>. not sure if there is another way around this

danciaclara:
will be fixed by https://github.com/appsmithorg/appsmith-docs/pull/904

PR #891, opened 2023-02-20
fix: Update migrate-to-helm-chart-v2-ce backup file name

george-paw:
The backup tar must have timestamp it in, otherwise `backup` would not work.


sum35h:
Nice catch, @george-paw. Thanks : )

sum35h (approved):

PR #893, opened 2023-02-21
fix: Notes for lack of HS256 support

infinitetrooper:
Adding a note for the lack of support of the HS256 algorithm in OIDC token signing.

I've added the note in each IDP page subtly because this is an edge case setup. I'm worried that users might get unnecessarily distracted if we show the note promptly or on the OIDC page.


danciaclara (dismissed):

sharat87 (changes requested):
Slight rewording:

> Please note, verifying tokens signed with the `HS256` algorithm isn't supported.

It's not that _signing_ is not supported, it's _verifying the signature_ that's not supported. The IdP will sign the token, Appsmith just verifies that signature. 🙂

danciaclara (approved):

sharat87 (approved):

PR #903, opened 2023-02-23
fix: Changes to Authentication docs

infinitetrooper:
1. Changes the entry point to authentication docs. It is part of Admin settings and the current entry point inside the instance configuration is confusing. Other Admin settings feature docs are in Advanced concepts. I would like to hear the reviewer's thoughts on this.
2. Updated the JWT documentation to focus on the important information at the top of the page.


infinitetrooper:
Will break this up into multiple PRs

PR #953, opened 2023-03-06
docs: update installation docs for appsmith-ee helm

pratapaprasanna:


website/docs/getting-started/setup/installation-guides/kubernetes/migrate-to-be-chart.md
```diff
@@ -136,7 +136,7 @@ To ensure that the Business Edition Helm chart runs, you need to make some chang
      APPSMITH_KEYCLOAK_DB_NAME: "keycloak"
    ```
 
-## Install v2 Helm chart 
+## Install v3 Helm chart
```
jnikhila, line 139: What about the installations which are already on v2?

jnikhila: @pratapaprasanna - Please confirm.

jnikhila (changes requested):

website/docs/getting-started/setup/installation-guides/kubernetes/migrate-to-be-chart.md
pratapaprasanna: - This will also work and we can avoid the confusion around using a specific version
- We have a new version v3 which fixes few issues

jnikhila: Does this mean that the updates to the command will always pull the latest version, and should the heading be v3 in that case?

pratapaprasanna: Yes, that makes more sense rather than confusing people with versions we can just remove it

danciaclara (dismissed):

jnikhila (approved):

PR #960, opened 2023-03-07
fix: (showAlert) added duration note, updated media

spark-c:
* Added note that `showAlert` timer is 5 seconds and cannot be modified.
* Replaced outdated .gif with an updated .png image

Resolves [appsmithorg/appsmith#8417](https://github.com/appsmithorg/appsmith/issues/8417)


danciaclara (commented):
@spark-c as Saptami suggested, did we check if there is a workaround to enable users to display message for shorter duration?

> Suggest consulting a developer on a deeper know how on this to provide better clarity on how this function can be optimally used.

danciaclara (approved):

PR #987, opened 2023-03-13
fix: Added note for renaming resources in Git

infinitetrooper:


AnaghHegde (dismissed):

website/docs/advanced-concepts/version-control-with-git/commit-and-push.md
```diff
@@ -21,3 +21,6 @@ In case the remote counterpart of your current branch has some commits that are
 #### Merge Conflicts
 
 If there are any merge conflicts, you will have to resolve them manually on the repository. Once you fix the conflicts, you can try pulling the changes again.
+
+#### General notes
+1. Renaming a resource is the same as deleting the old resource and creating a new one in the Git file system. For ex, when you rename a query, you may see `2 queries modified` while trying to commit, the 2 changes are the old query being deleted and the new one being created.
```
danciaclara, line 26:
```suggestion
1. Renaming a resource is the same as deleting the old resource and creating a new one in the Git file system. For example, when you rename a query, you may see `2 queries modified` while trying to commit; the 2 changes are the old query being deleted and the new one being created.
```

danciaclara (changes requested):

danciaclara (approved):

PR #1006, opened 2023-03-15
docs: Add Easypanel as a deployment method

berksmbl:


danciaclara:
@sum35h @sharat87 can you check this, please?

berksmbl:
It's unfortunate to hear that. EasyPanel is a very practical and user-friendly control panel, and it's also included in documentation for many well-known applications. It would have been helpful for your users to be able to install it with just one click. I find it strange that you didn't include such a direction when there are about 10 alternative installation guides available.

PR #1010, opened 2023-03-16
feat: adding JS Object variable

Pranay105:
- added JS object variable with a simple example.
- a complex example to be added later
-  Not be merged right now.


website/docs/core-concepts/writing-code/javascript-editor-beta/README.md
```diff
@@ -46,6 +46,39 @@ You can call the functions defined in a JS Object by using the notation `{{ JS_O
 The JS Objects defined are available across APIs, Queries, or other JS Objects defined for **a particular page,** and have a **page-level access** and are **not** accessible **across pages**.
 :::
 
+### JS Object variable as local state
```
danciaclara, line 49: This shouldn't be under **Calling a JS Object function** It should be under **Create JS Object** under a subheading called **JS Object variables** Under that, show how to define and access the variables.


For local state, this should also be touched in the section where we show reactive programming to update widgets programmatically.

danciaclara (changes requested):
In this page

danciaclara:
@bharath31 I thought it would be nice to have it there before showing users how to display the function response in sync fields.

danciaclara (approved):

PR #1011, opened 2023-03-16
fix: Notes on git branch protections

infinitetrooper:
Note for errors caused by remote branch protections.


infinitetrooper:
@danciaclara  - one Vale suggestion didn't make sense so, left it as is.

danciaclara (approved):

PR #1076, opened 2023-03-31
chore: Mention transform data under run a query

Pranay105:
- Added a link to transform data section under Run a query section

### DIFF
--- a/website/docs/core-concepts/data-access-and-binding/querying-a-database/README.md
+++ b/website/docs/core-concepts/data-access-and-binding/querying-a-database/README.md
@@ -49,7 +49,7 @@ All query results are stored in the [data](/reference/appsmith-framework/query-o
 {{ Query1.data }}
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/core-concepts/data-access-and-binding/querying-a-database/README.md
```diff
@@ -49,7 +49,7 @@ All query results are stored in the [data](/reference/appsmith-framework/query-o
 {{ Query1.data }}
 ```
 
-DB Queries fetch data in an array of objects where each object is a row, and each key in the object is a column.
+DB Queries fetch data in an array of objects where each object is a row, and each key in the object is a column. However, each widget property supports a specific data type. In that case, you can [transform the array data](/core-concepts/data-access-and-binding/displaying-data-read#transform-data) to a data type that's valid for the respective property.
```
danciaclara, line 52:
```suggestion
Queries fetch data in an array of objects where each object is a row, and each key in the object is a column. However, each widget property supports a specific data type. You can [transform the data](/core-concepts/data-access-and-binding/displaying-data-read#transform-data) from the query response to match the expected datatype in the property.
```

danciaclara (changes requested):

danciaclara (approved):

Issue #1075, opened 2023-03-31
[Docs]: App Navigation v1


dhruvikn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Engineering Ticket Link

https://github.com/appsmithorg/appsmith/pull/19312

### Release Date

3 April 2023

### Release Number

NA

### Loom video

Video here in the PR - https://github.com/appsmithorg/appsmith/pull/19312

### Test plan/cases

https://www.notion.so/appsmith/Test-Plan-a7883ae4980d470690de5c62a41dd168

somangshu:
@jnikhila @danciaclara can we close this task

somangshu:
@jnikhila @danciaclara can we close this task

PR #1082, opened 2023-04-03
feat: Add docs for app navigation

dhruvikn:
First draft for #1075


website/docs/core-concepts/building-ui/designing-an-application/app-navigation.md
```diff
@@ -0,0 +1,87 @@
+# App Navigation
+
+We have introduced new navigation styles with better a default navigation - Top (Stacked), a variant for Top (Inline), and a collapsible Sidebar. You can now configure your app's navigation by opening the navigation settings tab inside the app settings pane and observe how your app with the selected navigation settings will look side by side as you change them.
```
somangshu, line 3: Some grammatical mistakes here!

somangshu: We should do a better job defining the diff b/w Top (Stacked), a variant for Top (Inline)

somangshu: For devs, its worth talking about the live preview mode as well!

website/docs/core-concepts/building-ui/designing-an-application/app-navigation.md
```diff
@@ -0,0 +1,87 @@
...
+
+<figure>
+  <img src="/img/app-navigation-new-default.png" alt="The new default navigation - Top (Stacked)"/>
+  <figcaption align="center"><i>The new default navigation - Top (Stacked)</i></figcaption>
+</figure>
+
+## Navigation settings
+
+:::info
+Configuring the navigation is an application-level change.
```
somangshu, line 13: is application-level good enough to explain what it mean, maybe saying that `the scope of this change remains limited to the application where you set this` be better?

website/docs/core-concepts/building-ui/designing-an-application/app-navigation.md
```diff
@@ -0,0 +1,87 @@
...
+| Show navbar      | Toggles the visibility of the navbar. If turned off, the navbar won't be visible on the deployed app. | On      | -                                                                      |
+| Orientation      | Changes if the navbar should be on top or side.                                                       | Top     | Top and Side. [Read more here.](app-navigation.md#orientation)         |
+| Variant          | Only available when the orientation is set to Top. Available options are Stacked and Inline.          | Stacked | Stacked and Inline. [Read more here.](app-navigation.md#variant)       |
+| Background color | Changes the background color of the navigation.                                                       | Light   | Light and Theme. [Read more here.](app-navigation.md#background-color) |
+| Show sign in     | Toggles the visibility of the sign-in button when someone who is not logged in views the app.         | On      | -                                                                      |
```
somangshu, line 38: With these prop should we also add the scope, i.e. where the change is going to be visible

website/docs/core-concepts/building-ui/designing-an-application/app-navigation.md
```diff
@@ -0,0 +1,87 @@
...
+</figure>
+2. **Side** - Shows a collapsible sidebar on the left side of the app.
+<figure>
+  <img src="/img/app-navigation-side.png" alt="Side navigation"/>
+  <figcaption align="center"><i>Side navigation</i></figcaption>
+</figure>
+
+#### Variant
+
+This option is inly available when the orientation is set to Top.
```
somangshu, line 57: inly > only

somangshu (commented):
Small suggestions, otherwise LGTM

website/docs/core-concepts/building-ui/designing-an-application/app-navigation.md
dhruvikn: Can you explain a bit more about what the scope means here? Can you please give an example? I didn't quite get you.

somangshu: Sure, my suggestion is to explicitly call out the scope for a function. 
ex: `Show sign in` is only going to take effect on the single app having this config. 
If it makes sense we can use this as an opportunity to show the advantage for business editions users

dhruvikn: Gotcha, all of these options have an application-wide scope, which we have earlier mentioned and I'm defining it a bit more clearly as you suggested in the [above comment](https://github.com/appsmithorg/appsmith-docs/pull/1082#discussion_r1155830162). Would that complete the purpose of sharing the scope?

somangshu (approved):
LGTM

jnikhila:
For now, let's close the PR. We can revisit it when we start working on the Build UI section. CC: @danciaclara

dilippitchika:
@danciaclara @jnikhila this feels confusing. I get why you might not want to prioritise it due to other tasks, but closing our PR means @dhruvikn will have to later revisit this and help you write the documentation. 

But for everyone who asks how to setup a side nav, why should support respond with a sample app or a video, they should respond with this article instead. We are also adding features like allowing people to change logos, using a minimal side nav etc. I think the logo upload flow will go into release next week. 

Can you give me a date when you will prioritise this and when you think this will be available?

PR #1105, opened 2023-04-07
fix: Date type column returns an error

harshilp24:


website/docs/reference/widgets/table/column-settings.md
```diff
@@ -50,7 +50,13 @@ The checkbox column type supports [inline editing](/reference/widgets/table/inli
...
+The date column type allows you to set up custom formatting options for date and time information. You can format and display the date using the Date Format and Display Format properties.
+
+* The **Date Format** property specifies how incoming date data should be interpreted. For example, lets say that the incoming date data is in the format `YYYY-MM-DD HH:mm` but the Date Format property is set to `DD/MM/YYYY`. In this case, the app would not be able to properly interpret the date information and would display an evaluation error.
+
+* The **Display Format** property specifies how the date information should be displayed to the user. For example, if the incoming date data is in the format `YYYY-MM-DD` but the Display Format property is set to `DD/MM/YYYY`, the date information would be displayed to the user in the desired format of `DD/MM/YYYY`.
+
+It's important to ensure that both properties are set correctly to handle and display date and time information in your app. For more information on handling dates, read about the [Datepicker widget](/reference/widgets/datepicker).
```
danciaclara, line 59: I don't think they need to refer to datepicker widget

website/docs/reference/widgets/table/column-settings.md
```diff
@@ -50,7 +50,13 @@ The checkbox column type supports [inline editing](/reference/widgets/table/inli
 
 #### Date
 
-The date column type displays formatted date and time information. For more information on handling dates, read about the [Datepicker widget](/reference/widgets/datepicker).
+The date column type allows you to set up custom formatting options for date and time information. You can format and display the date using the Date Format and Display Format properties.
+
+* The **Date Format** property specifies how incoming date data should be interpreted. For example, lets say that the incoming date data is in the format `YYYY-MM-DD HH:mm` but the Date Format property is set to `DD/MM/YYYY`. In this case, the app would not be able to properly interpret the date information and would display an evaluation error.
```
danciaclara, line 55:
```suggestion
* The **Date Format** property specifies the date format of the incoming data specified in the **Computed Value** property. The date should be specified in a format that can be parsed correctly. 

For example, if the incoming date is in the format `YYYY-MM-DD HH:mm` but the **Date Format** property selected is **DD/MM/YYYY**, it is not able to parse the date and displays 'Invalid date' in the column. In this case, you can fix it in two ways. 
* Update the option in the **Date Format** property to match the format in the **Computed Value** property. In this case, select formats like **YYYY-MM-DD**, **YYYY-MM-DD HH:mm**, YYYY-MM-DD hh:mm:ss, etc.
* Transform the date in the **Computed Value** property using `moment().format()` to match the one in the **Date Format** property. 
```

danciaclara: The explanation should be clearer so that the user understands. I had to go through the loom videos to get the point.

website/docs/reference/widgets/table/column-settings.md
```diff
@@ -50,7 +50,13 @@ The checkbox column type supports [inline editing](/reference/widgets/table/inli
...
+
+* The **Display Format** property specifies how the date information should be displayed to the user. For example, if the incoming date data is in the format `YYYY-MM-DD` but the Display Format property is set to `DD/MM/YYYY`, the date information would be displayed to the user in the desired format of `DD/MM/YYYY`.
```
danciaclara, line 57:
```suggestion
* The **Display Format** property specifies how the date information should be displayed to the user. For example, if the incoming date is in the format `YYYY-MM-DD` but the **Display Format** property is set to `DD/MM/YYYY`, the date information would be displayed to the user in the desired format of `DD/MM/YYYY`.
```

danciaclara (changes requested):

website/docs/reference/widgets/table/column-settings.md
danciaclara: remove both those lines

danciaclara (approved):

PR #1124, opened 2023-04-12
chore: linter and grammar fixes iteration 2

jnikhila:


website/docs/learning-and-resources/how-to-guides/how-to-get-container-logs.md
```diff
@@ -34,7 +34,7 @@ zip -r $targetZipFile "$stacksPath/logs/$service"
 ## Docker in remote servers
 
 - SSH into the remote server and note the absolute path of the stacks directory.
-- If you don’t remember the path use the same command above to locate it
+- If you don’t remember the path use the same command to locate it
```
danciaclara, line 37: it will be difficult for the user to know which same command. Also add 'above' to accepted words list so the linter doesn't flag it always.

```suggestion
- If you don’t remember the path, use the `docker inspect` command as shown in the above section to locate it.
```

danciaclara (changes requested):

website/docs/learning-and-resources/how-to-guides/how-to-get-container-logs.md
jnikhila: done

danciaclara (approved):

PR #1145, opened 2023-04-18
fix: table rehaul

harshilp24:


website/docs/reference/widgets/table/README.md
```diff
@@ -2,305 +2,316 @@
...
 
-### Transform table data
+```js
+{{fetchData.data}}
+```
+The Table widget displays all the data from the query. 
 
-Some API / Query responses might have deeply nested, unnecessary, or unhelpfully formatted fields. These can be transformed to fit your needs by parsing and processing the data with JavaScript. The JS [`map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) function is highly recommended for processing raw data from your queries to fit your table correctly.
+---
+**Example 2:** if the data retrieved from the query is not in the desired format, you can use JavaScript to transform it before passing it to the Table widget.
```
danciaclara, line 64: The current api example is more useful. Can we keep that as discussed during the review call?

website/docs/reference/widgets/table/README.md
```diff
@@ -2,305 +2,316 @@
...
 
-1. Enable the server-side pagination property in the table
-2. Call the API / query via the **onPageChange** event
-3. Configure pagination in the API / query using the [Offset](#offset-based-pagination) or [Key-based](#key-based-pagination) pagination method.
+ Offset-based pagination works by using the page number and size to calculate the offset of records to fetch from a database or API. 
 
-:::tip
-Turning on **Server side pagination** also enables the **Total records** property. This property is useful for helping to control the page buttons in the table header.
-:::
+To implement offset-based pagination in a `user's` database, reference the `pageOffset`, `pageNo`, and `pageSize` property values using curly braces in the query.
```
danciaclara, line 100: Need to rewrite this as discussed on call

website/docs/reference/widgets/table/README.md
```diff
@@ -2,305 +2,316 @@
...
+ 
+</Tabs>
 
-Now when `sendNewData` succeeds, your table automatically refreshes itself.
 
-### Update table data in real time
+## Server-side searching
 
-If you want to update data in the table periodically without requiring users to trigger the refreshes, you can use the `setInterval` function.
+When **Allow Searching** property is on, the table header includes a search bar to find matching records. Server-side search lets you request specific records from the server using search terms.
```
danciaclara, line 148: this needs to be rewritten as discussed on call

website/docs/reference/widgets/table/column-settings.md
```diff
@@ -1,184 +1,165 @@
-# Column Settings
```
danciaclara, line 1: Main comments for this page were:

- Remove Edit Table section
- Include important information from the current doc in Freeze column

danciaclara (changes requested):

danciaclara (approved):

PR #1157, opened 2023-04-20
feat: lets encrypt warn for high risk domain name

sum35h:
Added a warn block to tell users that Let's encrypt does not support cert provisioning for ephemeral domains since they are high risk.
ex: ec2-xx-xx-xx-xx.ap-south-1.compute.amazonaws.com


website/docs/getting-started/setup/instance-configuration/custom-domain/README.md
```diff
@@ -25,6 +25,10 @@ Appsmith uses [Let's Encrypt](https://letsencrypt.org) to generate and maintain
 * [Admin settings](#admin-settings)
 * [Environment Variables](#environment-variables)
 
+:::caution Attention
+Let's Encrypt does not support provisioning a certificate for ephemeral domain names since they are high-risk.
```
jnikhila, line 29: @sum35h - What about users who are using ephemeral domain names? How can they use SSL?

jnikhila (changes requested):

website/docs/getting-started/setup/instance-configuration/custom-domain/README.md
sum35h: Users should never use ephemeral domain names.
They are usually provided by cloud providers by default.
They keep changing and someday someone else will own them, so it is never recommended.
Users have to buy a valid domain name and create an entry there.

jnikhila (approved):

PR #1191, opened 2023-05-03
fix: to trigger analysis only for prod in scarf and smartlook

jnikhila:
Added env variable for production

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -184,23 +184,23 @@ const config = {
         darkTheme: darkCodeTheme,
       },
     }),
-  scripts: [
-    {
-      src:
-        '/scripts/intercomSettings.js',
-      async: false,
-    },
-    {
-      src:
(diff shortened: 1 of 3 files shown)


tejasahluwalia:
Could we add the condition in the `docusaurus.config.js` instead? 

Like this -
```
...
    scripts: [
		process.env.VERCEL_ENV === "production" && {
			src: "/scripts/intercomSettings.js",
			async: false,
		},
		process.env.VERCEL_ENV === "production" && {
			src: "/scripts/smartlook.js",
			async: false,
		},
		process.env.VERCEL_ENV === "production" && {
			src: "/scripts/analyticsEvents.js",
			defer: true,
		},
	],
...
```
Because now we're getting an error in the console in the deploy preview since `process.env.VERCEL_ENV` is only available during build step.
image

jnikhila:
> 

Done.

tejasahluwalia (approved):

PR #1262, opened 2023-05-17
fix: updates to installation instructions for Kubernetes

jnikhila:
## Checklist
I have:
- [x] linked to sample apps when relevant
- [x] minimized the callouts and added only when necessary
- [x] added the `queryString` parameter to the Tabs (if used)
- [x] masked PII in images. For example, login credentials, account details, and more
- [x] added images only when necessary
- [x] deleted the images that are no longer used for the updated pages in the PR
- [x] used the `<figure/>` tag instead of a markdown representation for images
- [x] added the `<figcaption/>` tag to add a caption to the image
- [x] added the `alt` attribute in the `` tag

## Updates
Below are the changes for this PR:
- moved prerequisites under each tab
- removed Minikube from Business Edition tab
- removed Minikube from Configure High Availability tab

### DIFF
--- a/1
+++ b/1
@@ -1,0 +1,6 @@
+Merge branch 'main' into fix-kubernetes-installation-steps
+# Please enter a commit message to explain why this merge is necessary,
+# especially if it merges an updated upstream into a topic branch.
+#
+# Lines starting with '#' will be ignored, and an empty message aborts
+# the commit.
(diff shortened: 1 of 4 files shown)


danciaclara (commented):
This format looks confusing. Can we have pre-requisites outside as before. its just one tab of Minikube that is different and hidden on Business. Can we just mention that HA will not be suppored for this method and that if they want HA they should use one of the other three.

jnikhila:
> This format looks confusing. Can we have pre-requisites outside as before. its just one tab of Minikube that is different and hidden on Business. Can we just mention that HA will not be suppored for this method and that if they want HA they should use one of the other three.

Based on our discussion, updates are done.

danciaclara (dismissed):

danciaclara (approved):

pratapaprasanna (approved):

PR #1394, opened 2023-06-22
fix: add callout to mention env variable takes precedence over Admin settings

Pranay105:
## Checklist
I have:

### DIFF
--- a/website/docs/getting-started/setup/instance-configuration/README.mdx
+++ b/website/docs/getting-started/setup/instance-configuration/README.mdx
@@ -26,6 +26,9 @@ If you don’t see the “Admin Settings” option on the left sidebar, you’re
   <figcaption align = "center"><i>Access admin settings on the left panel under workspaces</i></figcaption>
 </figure>
 
+:::info
+When configuring an Appsmith instance, it's important to note that if you choose to set configurations using [environment variables](#environment-variables), those values take precedence over any values set in the Admin Settings.
+:::
 
 ### Add super admins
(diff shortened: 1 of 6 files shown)


website/docs/getting-started/setup/instance-configuration/README.mdx
```diff
@@ -26,6 +26,9 @@ If you don’t see the “Admin Settings” option on the left sidebar, you’re
   <figcaption align = "center"><i>Access admin settings on the left panel under workspaces</i></figcaption>
 </figure>
 
+:::info
+When configuring an Appsmith instance, it's important to note that if you choose to set configurations using [environment variables](#environment-variables), those values take precedence over any values set in the Admin Settings.
```
danciaclara, line 30:
```suggestion
If you have set values using [environment variables](#environment-variables) for your instance, those values take precedence over values specified in the Admin Settings UI.
```

danciaclara (changes requested):

website/docs/getting-started/setup/instance-configuration/email/README.md
```diff
@@ -45,6 +45,10 @@ Be sure to double-check your configuration if you find that you're able to send
 
 ## Configure using admin settings
 
+:::info
+If you have set values using [environment variables](#environment-variables) for your instance, those values take precedence over values specified in the Admin Settings UI.
```
danciaclara, line 49: this link in this callout is not working. Just change the headings in the page to match the other pages and should solve the issue

danciaclara (changes requested):
@Pranay105 this is good but I think the cross reference of environment variables is not working in the

danciaclara (approved):

PR #1461, opened 2023-07-12
fix: updating Query object

Pranay105:
## Checklist
I have:
- [x] minimized the callouts and added only when necessary
- [x] added the `queryString` parameter to the Tabs (if used)
- [x] added images only when necessary

### DIFF
--- a/website/docs/reference/appsmith-framework/query-object.mdx
+++ b/website/docs/reference/appsmith-framework/query-object.mdx
@@ -12,7 +12,37 @@ The query object contains the parameters required to run queries and access the 
 
 ## Properties
 
-Query object has the following properties that you can use to reference and control data related to your query.
+The query object has the following properties that you can use to reference the query's response or check the status of the query.
+
+### data `Array`
+
+<dd>
+
+`data` is a read-only property that contains the response body from the last successful execution of this query. 
+
+```
+{{QUERY_NAME.data}}
+```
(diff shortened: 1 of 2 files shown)


website/docs/reference/appsmith-framework/query-object.mdx
```diff
@@ -34,14 +41,13 @@ This function returns a JavaScript **promise**, which can be used to handle asyn
...
 ```
+It can't be used in [synchronous fields](/core-concepts/writing-code/workflows#display-data-from-async-js-function).
 
-To learn more about chaining actions to create complex workflows, see [complex workflows](/core-concepts/writing-code/workflows#complex-workflows).
-
-### Pass parameters to run() {#pass-parameters-to-run}
+</dd>
 
-In some cases, like [running a query inside a loop](/connect-data/how-to-guides/how-to-pass-params-to-an-api), parameters may need to be passed to the query with values contextual to the execution.
+#### Pass parameters to run()
```
danciaclara, line 48: this was supposed to be removed and linked to the how to the guide if the example is the same

danciaclara (changes requested):
Let's segregate the properties 

data (Array)
isLoading(boolean) - missing in docs
reponseMeta - Unknown according to platform- so check with FE pod what data type this is

and methods
run()
clear() - this is a function not an object

website/docs/reference/appsmith-framework/query-object.mdx
Pranay105: The example is not same. That's why I kept it. Forgot to mention in the message.

danciaclara (approved):

PR #1500, opened 2023-07-21
fix: Chart Rehaul review-fix

harshilp24:
## Checklist
I have:

### DIFF
--- a/website/docs/reference/widgets/chart.mdx
+++ b/website/docs/reference/widgets/chart.mdx
@@ -124,11 +124,11 @@ Allows you to add data for Custom charts. You need to specify `type` for chart t
 }
 ```
 
-* **Type :** The type is used to define the chart type recognized by the FusionCharts. FusionCharts provides an extensive range of chart variants, with almost 100+ options to choose from. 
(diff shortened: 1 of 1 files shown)


website/docs/reference/widgets/chart.mdx
```diff
@@ -124,11 +124,11 @@ Allows you to add data for Custom charts. You need to specify `type` for chart t
 }
 ```
 
-* **Type :** The type is used to define the chart type recognized by the FusionCharts. FusionCharts provides an extensive range of chart variants, with almost 100+ options to choose from. 
+* **Type :** The type is used to define the chart type recognized by the FusionCharts, written in JavaScript. FusionCharts provides an extensive range of chart variants, with almost 100+ options to choose from. For example: `"type": "column2d"`
```
jnikhila, line 127: Here you have to mention that the JavaScript alias should be used, and for datasource the JSON format.
Screenshot 2023-07-21 at 12 10 04 PM

jnikhila (changes requested):

jnikhila (approved):

PR #1502, opened 2023-07-21
fix: Form review

harshilp24:
## Checklist
I have:

### DIFF
--- a/website/docs/reference/widgets/form.md
+++ b/website/docs/reference/widgets/form.md
@@ -19,7 +19,7 @@ These properties are customizable options present in the property pane of the wi
 
 Controls the visibility of the widget. If you turn off this property, the widget would not be visible in View Mode. Additionally, you can use JavaScript by clicking on **JS** next to the **Visible** property to conditionally control the widget's visibility.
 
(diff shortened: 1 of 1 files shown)


website/docs/reference/widgets/form.md
```diff
@@ -19,7 +19,7 @@ These properties are customizable options present in the property pane of the wi
 
 Controls the visibility of the widget. If you turn off this property, the widget would not be visible in View Mode. Additionally, you can use JavaScript by clicking on **JS** next to the **Visible** property to conditionally control the widget's visibility.
 
-For example, if you want to make the widget visible only when the user selects "Yes" from a Select widget, you can use the following JavaScript expression: 
+For example, if you want to make the widget visible only when the user selects `Yes` from a Select widget, you can use the following JavaScript expression:
```
jnikhila, line 22:
```suggestion
For example, if you want to make the widget visible only when the user selects _Yes_ from a Select widget, you can use the following JavaScript expression: 
```

jnikhila (changes requested):

jnikhila (approved):

PR #1560, opened 2023-08-02
docs: Initial draft for doc on configuring HTTP/HTTPS proxy

sharat87:
Preview at https://appsmith-docs-git-docs-http-proxy-get-appsmith.vercel.app/getting-started/setup/instance-configuration/http-proxy.

## Checklist
I have:


website/docs/getting-started/setup/instance-configuration/http-proxy.md
```diff
@@ -0,0 +1,23 @@
...
+---
+# HTTP/HTTPS Proxy
+
+Appsmith supports running behind a forward proxy. This is usually useful when Appsmith is run on a server that's behind a corporate firewall and needs to access the internet through a proxy.
+
+## How to configure a proxy
+
+As is standard for configuring a proxy for most applications, Appsmith respects any `HTTP_PROXY` and/or `HTTPS_PROXY` env variables that are set. Fox extra compatibility, Appsmith also respects the lower-case versions, `http_proxy` and `https_proxy`.
+
+For example, adding the following to the `stacks/configuration/docker.env`:
```
jnikhila, line 12: Are these configuration variables also available for Kubernetes - `values.yaml` file?

sharat87: Yes, they're environment variables, they can be set anywhere where environment variables can be set. This is just an example that I put in, happy to rephrase. 🙂

website/sidebars.js
```diff
@@ -208,6 +208,7 @@ const sidebars = {
                     'getting-started/setup/instance-configuration/custom-domain/custom-ca-root-certificate',
                   ],
                 },
+                'getting-started/setup/instance-configuration/http-proxy',
```
jnikhila, line 211: @sharat87 - Please confirm - The Proxy will be used in tandem with the custom domain configuration. Will it make sense to have this page added under the custom domain?

jnikhila: CC: @danciaclara

sharat87: > The Proxy will be used in tandem with the custom domain configuration.

Not necessarily. Having or not having a custom domain isn't related to having to use a `HTTP_PROXY` or not. They're unrelated.

website/docs/getting-started/setup/instance-configuration/http-proxy.md
```diff
@@ -0,0 +1,34 @@
...
+- A self-hosted Appsmith instance. See the [installation guides](/getting-started/setup/installation-guides) to set up your Appsmith instance.
+
+## Configure HTTP/HTTPS proxy
+
+Follow these steps to configure HTTP/HTTPS proxy for your self-hosted instance:
+
+1. Go to your Appsmith instance configuration file, such as the `docker.env` file for Docker or the `values.yaml` file for Kubernetes.
+2. Use any one of the below pair of environment variables to set proxy:
+
+ a. `HTTP_PROXY` and/or `HTTPS_PROXY`
```
danciaclara, line 20: Can we write

a.  `HTTP_PROXY` or `http_proxy` 
b. `HTTPS_PROXY` or `https_proxy`

instead of the other way around

website/docs/getting-started/setup/instance-configuration/http-proxy.md
```diff
@@ -0,0 +1,34 @@
...
+Follow these steps to configure HTTP/HTTPS proxy for your self-hosted instance:
+
+1. Go to your Appsmith instance configuration file, such as the `docker.env` file for Docker or the `values.yaml` file for Kubernetes.
+2. Use any one of the below pair of environment variables to set proxy:
+
+ a. `HTTP_PROXY` and/or `HTTPS_PROXY` 
+
+ b. `http_proxy` and/or `https_proxy`
+
+3. Add the chosen environment variables as shown below:
```
danciaclara, line 24: Write this as an example for step 2.  it does not appear to be a different step

website/docs/getting-started/setup/instance-configuration/http-proxy.md
```diff
@@ -0,0 +1,34 @@
...
+ ```bash
+ HTTP_PROXY=http://1.2.3.4:8080
+ HTTPS_PROXY=http://1.2.3.4:8080
+ ```
+4. Save the changes and restart the Appsmith instance.
+5. To use an HTTPS proxy, you also need to add a trusted custom Certificate Authority (CA) to Appsmith. See [Custom Root CA Certificate](/getting-started/setup/instance-configuration/custom-domain/custom-ca-root-certificate) to set up a certificate authority.
```
danciaclara, line 30: This looks more like a prerequisite to me

danciaclara (changes requested):

website/docs/getting-started/setup/instance-configuration/http-proxy.md
jnikhila: No, either the users can use HTTP_PROXY and HTTPS_PROXY or choose to use lower case. Also, they can choose to use either HTTP_PROXY or HTTPS_PROXY or both. @sharat87 - Please confirm this.

website/docs/getting-started/setup/instance-configuration/http-proxy.md
```diff
@@ -0,0 +1,34 @@
+---
+description: This page provides steps to configure HTTP/HTTPS proxy on your self-hosted instance.
+---
+
+# Configure Proxy
+
+Appsmith offers support for running behind a forward proxy, which is beneficial when deploying Appsmith on a server behind a corporate firewall, and allows internet access through a proxy server. This page provides instructions for configuring HTTP/HTTPS proxy on your self-hosted Appsmith instance.
```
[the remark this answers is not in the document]
jnikhila: why do we want to change it - `provides instructions` is also accurate.

website/docs/getting-started/setup/instance-configuration/http-proxy.md
```diff
@@ -0,0 +1,34 @@
+---
+description: This page provides steps to configure HTTP/HTTPS proxy on your self-hosted instance.
+---
+
+# Configure Proxy
```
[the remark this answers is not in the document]
sharat87: Let's keep to `Configure HTTP/HTTPS Proxy`, and not include the word `Server` please. Otherwise it sounds like we're about to talk about configuring the server itself. 🤔

website/docs/getting-started/setup/instance-configuration/http-proxy.md
sharat87: Sure, we can have it before, HTTPS proxies can be used without needing a custom CA, I think, but it's unusual to use it that way. And will most likely throw errors. I don't think we should do it as a prerequisite.

website/docs/getting-started/setup/instance-configuration/http-proxy.md
```diff
@@ -0,0 +1,34 @@
...
+
+Appsmith offers support for running behind a forward proxy, which is beneficial when deploying Appsmith on a server behind a corporate firewall, and allows internet access through a proxy server. This page provides instructions for configuring HTTP/HTTPS proxy on your self-hosted Appsmith instance.
+
+## Prerequisites
+
+- A self-hosted Appsmith instance. See the [installation guides](/getting-started/setup/installation-guides) to set up your Appsmith instance.
+
+## Configure HTTP/HTTPS proxy
```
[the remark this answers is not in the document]
danciaclara: Don't want too many configure words on the page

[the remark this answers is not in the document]
danciaclara: Also we are not configuring anything on Appsmith as such. Its is in the configuration file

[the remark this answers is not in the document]
jnikhila: The environment variable is part of the Appsmith Configuration, so we are configuring it on Appsmith.

website/docs/getting-started/setup/instance-configuration/http-proxy.md
jnikhila: @sharat87 - Does this mean that:
* Either this is more like a recommended setting, or this can serve as a troubleshooting guide to fix an error that a user might see.

sharat87: No, I don't think this should be a recommendation, since such a recommendation should come in from their company policy that required the proxy in the first place. I wouldn't presume to make a recommendation here.

I also don't think it should go in troubleshooting guide, cause, I don't want them to go see the error, then come back, look around in the troubleshooting, and then figure it out. This is common enough that we should frame it something like:

> When using an `HTTPS_PROXY`, you'll usually also want to add a custom CA to be trusted. If this is the case for your proxy, please check out this other document.

Or something similar.

danciaclara (approved):

Issue #697, opened 2022-12-21
[Docs]: Add feedback widget in Documenation


danciaclara:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Engineering Ticket Link

na

### Release Date

na

### Release Number

 na

### Use cases or user requests

In Gitbook we had a feedback widget with emojis where users could rate whether the page is useful. 
<img width="319" alt="Screenshot 2022-12-21 at 23 43 36" src="https://user-images.githubusercontent.com/32227135/208975248-4fdfea32-d506-47f4-b62b-2408ff2fb9ae.png">

This is very important information for us to track. Docusaurus has this feature planned in their roadmap, but until then, we should build this using Swizzle.
https://happyreact.com/blog/feedback-widget-in-docusaurus

jnikhila:
@danciaclara - The [link](https://happyreact.com/blog/feedback-widget-in-docusaurus) uses Docusaurus Footer. But as we are going to remove the footer as part of #519, we will create a new component to add this.

jnikhila:
Based [on this discussion](https://theappsmith.slack.com/archives/C048H8SEJ8K/p1680509736880889) - We'll need to evaluate the best possible way forward. Moving it back to Backlog ready pipeline.

PR #1662, opened 2023-08-24
feat: add doc feedback widget

jnikhila:
## Changes
- Added a feedback widget
    - Available on the Table of contents bar
    -  A 'No' will show a survey link to users to gather more info.

## Doc ticket
Fixes - #697

## Checklist
I have:
- [x] added the `alt` attribute in the `` tag


website/src/components/feedback/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+    // Update the feedback state after handling
+    setFeedback({
+      ...feedback,
+      helpful: value,
+    });
+  };
+
+  return (
+    <div className="feedback-widget-container">
```
tejasahluwalia, line 29: Using a `<form>` element would be semantic HTML. Will help with keyboard navigation, screen readers, etc.
Will have to change the handler and call it `onSubmit` for this to work. So leave it up to you if you want to change this now, or later if we add comments to the feedback as well.

website/src/theme/TOC/index.js
tejasahluwalia, line 1: Since its added to the TOC, we don't see the widget in mobile at all. Just checking if we are aware and okay with this.

website/src/css/custom.css
```diff
@@ -445,3 +445,48 @@ ul.circle > li:before {
...
+  border: 1px solid #ccc;
+  border-radius: 8px;
+  padding: 0.5em !important;
+  text-align: center;
+  width: 17%;
+  max-height: 25%;
+  margin-top: 2em !important;
+  margin-right: 1.5em !important;
+
+}
```
tejasahluwalia, line 460: removed `position: fixed` and set `width: 100%`

```suggestion
.feedback-widget-container {
  border: 1px solid #ccc;
  border-radius: 8px;
  padding: 0.5em !important;
  text-align: center;
  width: 100%;
  margin-top: 2em !important;
  margin-right: 1.5em !important;
}
```

tejasahluwalia (commented):
Let me know what you think of the comments I've added.

website/src/components/feedback/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const handleHelpfulChange = async (value) => {
+    // Send the feedback to Segment first
+    const feedbackJSON = generateFeedback(value);
+    await sendToSegment(feedbackJSON);
+
+    // Then open the Intercom widget if "No" is chosen
+    if (value === 'no') {
+      if (typeof Intercom !== 'undefined') {
+        Intercom('show');
```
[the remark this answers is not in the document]
jnikhila: This will be changed, as we'll use a survey here.

website/src/components/feedback/index.js
jnikhila: That's true. However, as there is no input we are capturing, and triggering segment as soon as the user chooses an option, I don't think we need a `<form/>` element here. Also, as the survey will cater to capturing detailed feedback comments, we are not adding comments to the feedback widget.

website/src/css/custom.css
jnikhila: position: fixed is needed so that the feedback widget is always available even if the page is scrolling. With position, width is also needed.

website/src/theme/TOC/index.js
jnikhila: This should be okay.

tejasahluwalia (approved):

tejasahluwalia (dismissed):

website/src/css/custom.css
```diff
@@ -488,4 +488,9 @@ body {
   border-color: #f86a2b;
   font-size: 0.8em;
   font-weight: 500;
+}
+
+.fixedToc {
+  position: fixed;
+  width: 17em;
```
tejasahluwalia, line 495: Fixed width value leads to it being cut off. Think it needs a `max-width: 15%` or something also
image

tejasahluwalia (changes requested):

danciaclara (approved):

PR #1801, opened 2023-10-11
fix: Removed SAML references from SCIM

infinitetrooper:
## Checklist
I have:

### DIFF
--- a/website/docs/advanced-concepts/user-provisioning-group-sync.mdx
+++ b/website/docs/advanced-concepts/user-provisioning-group-sync.mdx
@@ -48,9 +48,9 @@ This page shows how to configure Appsmith for SCIM to enable the automatic provi
 To complete the setup, go to your IdP application that is connected to Appsmith and follow the steps below:
 
 <Tabs queryString="scim">
-<TabItem label="Active Directory (SAML)" value="scim-ad-saml">
+<TabItem label="Active Directory" value="scim-ad">
 
-1. On the Active Directory homepage, click **Enterprise application** on the sidebar and open the SAML application that is already connected to Appsmith.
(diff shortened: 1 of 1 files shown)


website/docs/advanced-concepts/user-provisioning-group-sync.mdx
```diff
@@ -48,9 +48,9 @@ This page shows how to configure Appsmith for SCIM to enable the automatic provi
 To complete the setup, go to your IdP application that is connected to Appsmith and follow the steps below:
 
 <Tabs queryString="scim">
-<TabItem label="Active Directory (SAML)" value="scim-ad-saml">
+<TabItem label="Active Directory" value="scim-ad">
 
-1. On the Active Directory homepage, click **Enterprise application** on the sidebar and open the SAML application that is already connected to Appsmith.
+1. On the Active Directory homepage, click **Enterprise application** on the sidebar and open the application that is already connected to Appsmith.
```
jnikhila, line 53: Does the application have to be connected to Appsmith?

infinitetrooper: If they configured SSO first (which is most likely), then yes.

jnikhila: What if they don't, still SCIM works right? We should mention that if you have configured SSO, use the same connected app or create a new app on the provider.

infinitetrooper: SCIM connection works but, SSO won't so, it makes little sense to setup SCIM before SSO.

jnikhila (approved):

PR #1818, opened 2023-10-17
fix: Guide for fetch and filter GSheet data

### LINKED ISSUE #1816: [Docs]:Retrieve and transform data from Google sheets - Howto-Guide
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Documentation Link

_No response_

### Discord/slack/intercom Link

_No response_

### Describe the problem and improvement.

This ticket covers the documentation on how to fetch, filter, and transform Google sheet data in Appsmith.

rashmi-sahoo-git:
closes #1816 
## Checklist
I have:


sneha122:
@rashmi-sahoo-git Here are couple of feedback points, let me know what you think:
- Can we make it clear in the document, that when you create a new gsheet query, filter format will appear only after we select spreadsheet and sheet, because as a new user if I am following this document, initially filter format will be hidden and will only be seen once you select operation, spreadsheet and sheet. This will set the correct context and there wont be confusion for the user
- Same for cell range, we need to set the context before we start with the guide i.e. we need to select spreadsheet, sheet etc before filter options are shown
- Is it possible to add a small image of google sheet rows and columns to show the row number and column letter syntax? This is a very nit pick kind of thing, but I think would be very useful for someone who does not know much about it
- Cell_range_row and Cell_range_col are these default variables provided by appsmith to access cell range? or do we need to define it somewhere? If we need to define it and use, can we have a link to how we can define these and use it? Because for new user the concept of binding may be new and they might end up getting stuck at this phase
- What pagination offset does is not very clear from the documentation, can we please explain it in a better way if possible?

website/docs/connect-data/how-to-guides/filter-data-google-sheet.md
```diff
@@ -0,0 +1,84 @@
...
+  </iframe>
+</div>
+
+
+## Prerequisites
+- An app connected to a [Google Sheets](/connect-data/reference/querying-google-sheets) datasource.
+- A [Table](/reference/widgets/table) widget.
+- A query configured to fetch all the rows from a specific Google Sheet and bind to the Table widget.
+
+### Filter data using where clause
```
danciaclara, line 21:
```suggestion
### Using where clause
```

website/docs/connect-data/how-to-guides/filter-data-google-sheet.md
```diff
@@ -0,0 +1,84 @@
...
+   ```jsx
+   {{gender.text}}
+   ```
+4. In **[Sort By](/connect-data/reference/querying-google-sheets#sort-by)** enter the column name you want your data sorted by.
+   
+:::info
+If the **Smart JSON substitution** setting is enabled, Appsmith adds or removes quotation marks from the mustache binding `{{}}` as necessary to correctly cast them into JSON. You must manually format the JSON data if this setting is off. For a video guide on using this feature, see [How to Use Smart JSON Substitution](https://www.youtube.com/watch?v=-Z3y-pdNhXc).
+:::
+
+### Filter data using a cell range
```
danciaclara, line 43:
```suggestion
### Using a cell range
```

website/docs/connect-data/how-to-guides/filter-data-google-sheet.md
```diff
@@ -0,0 +1,84 @@
...
+   ```
+   Your fetched data still includes the column labels even if the column header row does not appear in your selection.
+   
+   For more information, see [Cell Range](/connect-data/reference/querying-google-sheets#cell-range).
+3. To dynamically bind the cell range using mustache syntax, use the following format where `Cell_range_row` and `Cell_range_col` are two [Text](/reference/widgets/text) widgets:
+
+   ```jsx
+   {{Cell_range_row.text}}:{{Cell_range_col.text}}
+   ```
+### Pagination
```
danciaclara, line 64:
```suggestion
### Server-side Pagination
```

danciaclara (changes requested):

danciaclara (approved):

PR #1821, opened 2023-10-18
fix: Fetch and filter data in SQL

harshilp24:
## Checklist
I have:

### DIFF
--- a/website/docs/connect-data/how-to-guides/fetch-and-filter-data-in-sql.md
+++ b/website/docs/connect-data/how-to-guides/fetch-and-filter-data-in-sql.md
@@ -1,0 +1,217 @@
+# Fetch and filter data in SQL
+
+This guide shows you how to read SQL data, apply filters, and display the results in widgets.
+
+
+## Prerequisites
+
+* App Connected to [PostgreSQL](/connect-data/reference/querying-postgres) Datasource
+* A Table widget to display data.
+
+## Display Data
+
+Follow these steps to perform a basic read operation and present the data within widgets. 
+
+1. In **Queries/JS**, click the **+** and create a query to retrieve the data using a simple SQL query like this:
+
+<dd>
+
+*Examples:* 
+
+```sql
+-- Retrieve All Columns from the "users" Table
+SELECT * FROM users;
+
+-- Select Specific Columns (id and name) from the "users" Table
+SELECT id, name FROM users;
+```
+
+</dd>
+
+2. To display the query result, use the mustache syntax `{{}}` in the widget property:
+
+
+<dd>
+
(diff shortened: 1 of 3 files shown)


website/docs/connect-data/how-to-guides/fetch-and-filter-data-in-sql.md
```diff
@@ -0,0 +1,255 @@
...
+
+
+
+```sql
+-- With prepared statements disabled
+SELECT * FROM public."features" where name ILIKE '%{{search_input.text}}%' order by created_at desc
+```
+
+
+Placing the `%`character within single quotes can lead to unexpected results and security vulnerabilities. It doesn't provide the same level of security and flexibility as the recommended format, making it more susceptible to SQL injection attacks.
```
sumitsum, line 157: I am not feeling confident about this line - `Placing the `%`character within single quotes can lead to unexpected results and security vulnerabilities.` - IMO, the security concern should not be because of where the quotes are placed - it should be only becuase prepared statements are not used - because without prepared statements it becomes vulnerable to SQL injection attacks.

website/docs/connect-data/how-to-guides/fetch-and-filter-data-in-sql.md
```diff
@@ -0,0 +1,255 @@
...
+
+This query uses the ANY keyword to match the `name` column in the `employees` table with both the `name` and `dept` values together in a structured array.
+
+
+</dd>
+
+
+## Handling IS Keyword in Prepared Statements
+
+When working with prepared statements, the IS keyword is not supported. Instead, you should rewrite your queries using the `=` operator.
```
sumitsum, line 225: Minor feedback: The `IS` keyword is supported for all databases except MySQL.

website/docs/connect-data/how-to-guides/fetch-and-filter-data-in-sql.md
```diff
@@ -0,0 +1,255 @@
...
+<dd>
+
+**Avoid**
+
+```sql
+-- With prepared statements disabled
+SELECT * FROM student WHERE pass IS {{true}};
+```
+
+Using the IS keyword in your queries with prepared statements can lead to syntax errors. It's essential to switch to the = operator to maintain query integrity and functionality.
```
sumitsum, line 252: `Using the IS keyword in your queries with prepared statements can lead to syntax errors. It's essential to switch to the = operator to maintain query integrity and functionality.` - As per my understanding this is not correct. The reason we don't support `IS` keyword is because of some bug in the MySQL database driver the we used to have.

danciaclara (approved):

PR #1887, opened 2023-11-22
feat: Open AI Datasource

harshilp24:
## Checklist
I have:


website/docs/connect-data/reference/open-ai.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### Messages
+
+<dd>
+
+Messages serve as input interactions between the user and the model. The roles include two types: **user** and **system**. You can create multiple messages of each type to make your conversation just the way you want. 
+
+* **user**: Input provided by the user to instruct or guide the model. For example, if you are using an Input widget to enter the prompt, you can use `{{Input1.text}}`.
+
+* **system**: The system message serves as a means to provide additional context, set guidelines, or convey the overall objective of the task. It helps shape the behavior of the model's responses. For example, you can use the system message to give personality to the responses or add task-specific instructions. For example, you can set system as:
```
dvj1988, line 84: System messages are usually added before the user message. Can you reorder this?

website/docs/connect-data/reference/open-ai.md
```diff
@@ -0,0 +1,222 @@
...
+
+*Options:*
+
+* **Float**: Returns the embeddings in float format.
+* **Base64**: Returns the embeddings encoded in Base64 format.
+
+</dd>
+
+
+You cannot directly bind the output to a widget, as this vector cannot be consumed on its own. Instead, you can store it in a specialized database called a Vector DB. Later, you can leverage these vectors for semantic search operations to retrieve related content based on the meaning of a query. For more information on embedding refer to the [OpenAI documentation.](https://platform.openai.com/docs/api-reference/embeddings/object)
```
dvj1988, line 153: I think we need to rephrase this.
> You cannot directly bind the output to a widget

The users can definitely bind it to a widget if they wish to but there is no point in doing so. This phrase sends a message that its is not possible.

I asked ChatGPT to rephrase this. 
```suggestion
While it is technically possible to bind the output of the embeddings API to a widget, doing so offers little practical value, as the vector representation is not meaningful for direct display in the UI. Instead, you can store it in a specialized database called a Vector DB. Later, you can leverage these vectors for semantic search operations to retrieve related content based on the meaning of a query. For more information on embedding refer to the [OpenAI documentation.](https://platform.openai.com/docs/api-reference/embeddings/object)
```

dvj1988 (commented):
Have just couple of comments. Rest looks good to me.

@bharath31 Please review this. You can check the vercel preview link.

danciaclara (dismissed):

danciaclara (approved):

PR #1986, opened 2023-12-14
feat: environment variables reference

jnikhila:
## Checklist
I have:


sharat87 (changes requested):
1. Can we use level-3 headers under level-2, instead of having level-4 under level-2 please?
2. I see `dd` tags without corresponding `dt` tags. Can we remove the `dd` tags since they don't seem to be necessary here?
3. All the env variables have a double underscore after `APPSMITH`. It should be one underscore only.

jnikhila:
> 1. Can we use level-3 headers under level-2, instead of having level-4 under level-2 please?
     _The heading level for environment variables is intentionally kept at level 4 to address formatting concerns. This ensures proper visual distinction between the main headings and subheadings, especially as all environment variable names are in uppercase. For comparison per se, I've updated one head so that we can visualize it. @danciaclara - let me know your thoughts too._
> 2. I see `dd` tags without corresponding `dt` tags. Can we remove the `dd` tags since they don't seem to be necessary here?
  _That is the formatting we are using across all reference docs._
> 3. All the env variables have a double underscore after `APPSMITH`. It should be one underscore only.
     _Fixed_

website/docs/getting-started/setup/environment-variables.md
```diff
@@ -0,0 +1,231 @@
...
+<dd>
+The secret key provided by Google reCAPTCHA that Google uses to verify the user input during communication between your server and the reCAPTCHA server. The secret is confidential and used in server-side code. Set this variable with the secret key provided by reCAPTCHA.
+</dd>
+
+
+## Database and session management
+
+Appsmith can connect to external providers for MongoDB and Redis. The associated environment variables ensure that Appsmith establishes a connection to external hosting rather than using the embedded MongoDB or Redis servers.
+
+#### `APPSMITH_MONGODB_URI`
```
jnikhila, line 170: @sharat87 - Could you please provide an example string that I can include here?

website/docs/getting-started/setup/environment-variables.md
```diff
@@ -0,0 +1,231 @@
...
+
+<dd>
+    Appsmith uses this variable to connect to an external MongoDB instance. Set it to the full MongoDB URI supplied by the hosting service. This enables Appsmith to store data in your MongoDB database, ensuring control and ownership of the application data.
+</dd>
+
+#### `APPSMITH_REDIS_URL`
```
jnikhila, line 176: @sharat87 - Could you please provide an example string that I can include here?

jnikhila:
@sharat87 - Added other envs. Available for final review. The heading level for all env variables is level 4. discussed this with @Dancia too. We agreed to keep it at level 4 to have a visual differentiation for headings and subheadings.

website/docs/getting-started/setup/environment-variables.md
```diff
@@ -0,0 +1,295 @@
...
+    This variable stores the Client ID provided by Google during OAuth 2.0 configuration. It's unique to your application, identifying you during the authentication process with Google's servers. Set this variable to the Google OAuth 2.0 Client ID available in your Google Cloud Platform console under the credentials section for your project.
+</dd>
+
+#### `APPSMITH_OAUTH2_GOOGLE_CLIENT_SECRET`
+
+<dd>
+    This variable holds the Client Secret provided by Google for OAuth 2.0. The secret is confidential and used in server-to-server communications. Assign this variable with the Google OAuth 2.0 Client Secret obtained from the Google Cloud Platform console where you configured your OAuth credentials.
+</dd>
+
+## Google OAuth
```
danciaclara, line 27:
```suggestion
## GitHub OAuth
```

danciaclara (changes requested):

danciaclara (approved):

PR #2034, opened 2024-01-04
fix: Fusion chart callout

harshilp24:
## Checklist
I have:

### DIFF
--- a/website/docs/reference/widgets/chart.mdx
+++ b/website/docs/reference/widgets/chart.mdx
@@ -144,7 +144,7 @@ See how to create a [Custom EChart](/build-apps/how-to-guides/create-custom-char
 <dd>
 
 :::caution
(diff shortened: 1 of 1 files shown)


website/docs/reference/widgets/chart.mdx
```diff
@@ -144,7 +144,7 @@ See how to create a [Custom EChart](/build-apps/how-to-guides/create-custom-char
 <dd>
 
 :::caution
-The support for Custom Fusion Charts ends on March 1st, 2024. Change the chart type to [Custom ECharts](#custom-echarts) to switch.
+Starting from March 1st, 2024, Custom Fusion Charts will display a watermark. To avoid this, consider transitioning to [Custom ECharts](#custom-echarts) for a watermark-free experience.
```
danciaclara, line 147:
```suggestion
Effective from March 1st, 2024, Custom Fusion Charts will display a watermark as the license will not be renewed. It is recommended to start using [Custom ECharts](#custom-echarts).
```

danciaclara (changes requested):

danciaclara (approved):

PR #2402, opened 2024-07-17
feat: Pass params values in JS Module Instance

harshilp24:
## Description 

Provide a concise summary of the changes made in this pull request
- 

## Pull request type

Check the appropriate box:

- [ ] Review Fixes
- [ ] Documentation Overhaul
- [x] Feature/Story
    - Link one or more Engineering Tickets
        * 
- [ ] A-Force
- [ ] Error in documentation
- [ ] Maintenance

## Documentation tickets

 Link to one or more documentation tickets:
 - 

## Checklist

From the below options, select the ones that are applicable:

- [ ] Checked for Grammarly suggestions.
- [ ] Adhered to the writing checklist.
- [ ] Adhered to the media checklist.
- [ ] Verified and updated cross-references or added redirect rules.
- [ ] Tested the redirect rules on deploy preview.
- [ ] Validated the modifications made to the content on the deploy preview.
- [ ] Validated the CSS modifications on different screen sizes.


website/docs/packages/reference/package.md
```diff
@@ -1,6 +1,9 @@
-# Package Settings
+# Module
+
+Modules in Appsmith are integral components of a Package tailored. They are reusable components containing queries and JSObjects that can yield results across multiple applications within the same workspace.
```
pedro-santos-rodrigues, line 3: Not sure if having this written in this way couldn't lead to some wrong expectation that a module can contain more than 1 query of JS Object

website/docs/packages/reference/query-module.md
```diff
@@ -24,7 +35,7 @@ The input property enables you to pass dynamic values from your app to the query
 You cannot edit the input name or query configuration from the App; you can only pass values to the available inputs.
```
pedro-santos-rodrigues, line 35: Suggestion: what is we complement this session with information that it's possible to change the values of inputs in runtime by passing different values in the .run() function? I've attached a video example that shows that it's possible and the value passed in runtime has precedence over the default value set in the module instance page.

[Video example](https://drive.google.com/file/d/14Nk5fuIhR_l_hNfupUjAhHDzUpPRHag9/view?usp=sharing)

website/static/img/query-module-instance.png
pedro-santos-rodrigues, line 1: In the text near this image, there is an example saying:

> If you have an input named `distinct_id`, you can pass data like this:

Can we update this image to have a better module name instead of QueryModule11 and the input with the same name that is used in the example? Also for the other screenshots I think it would be better if we have more real names, what do you think?

website/docs/packages/reference/query-module.md
```diff
@@ -70,56 +82,70 @@ When enabled, this property displays a confirmation modal each time before refre
 
 ## JS module
```
pedro-santos-rodrigues, line 83: What if we change this heading 2 to "JS module instance" so that it's more clear in the documentation the difference between JS module and JS module instance?

website/docs/packages/reference/query-module.md
```diff
@@ -1,6 +1,17 @@
...
+When you incorporate a module from the package into your application, you create a module instance. You can create multiple instances of the same package, each with different settings and configurations. Instances are named sequentially (e.g., `productutil1`, `productutil2`), allowing for customized functionalities.
+
+This page provides information about the module settings and properties available within the app, which allows you to configure the query and JS modules.
+
+
+<ZoomImage
+  src="/img/modules-diagram.png" 
+  alt=""
+  caption=""
+/>
```
pedro-santos-rodrigues, line 17: What if we change this heading 2 to "Query module instance" so that it's more clear in the documentation the difference between Query module and Query module instance?

website/docs/packages/reference/query-module.md
```diff
@@ -70,56 +82,70 @@ When enabled, this property displays a confirmation modal each time before refre
...
 
-<dd>
+#### Inputs
 
-This setting displays all the function names available in the JS module. 
 
+The input property allows you to pass dynamic values from your app to the JS module. The input parameters are only available if they are defined in the JS function. You cannot edit the input name or JS module configuration from the App; you can only pass values to the available inputs. 
 
+:::info
+When passing data from the app to the JS module using both **Inputs** and during **runtime** (e.g., `login.run({ email: <email>, password: <password> })`), the runtime parameters will take precedence over the values set in the UI.
```
pedro-santos-rodrigues, line 99: The information shown in this box is accurate for query modules but for JS Modules the syntax is different. When you want to pass arguments in runtime to a JS Function (login) that exist in a JS Module (authenticationUtils) you need to use the following syntax:

`authenticationUtils.login('email@domain.com', 'dsoi3$dfssn)`


Let me know if you have questions about the difference in both. I can help on clarifying.

website/docs/packages/reference/query-module.md
```diff
@@ -63,63 +79,80 @@ When enabled, this property allows the query to automatically execute each time
...
+}
+```
+
+```js
+//JS module instance
+email: 
+{{email_input.text}}
 
+passwordHash:
+{{pass_input.text}}
```
pedro-santos-rodrigues, line 130: Suggestion: what if the binding has a JSObject function that returns the hash based on the argument that is the pass_input.text? In that way, it seems more real the example and also demonstrates that people are also able to access JSObjects from those inputs.

pedro-santos-rodrigues (approved):
Added some feedback but approved because nothing blocker

jnikhila (approved):

PR #3028, opened 2026-07-31
docs: disable bundled DB subcharts in external database guides

sebastianiv21:
## Description

The three external-database guides (MongoDB, PostgreSQL/RDS, Redis) only document the connection URL variables for Kubernetes/Helm installs. With the chart-managed databases left enabled, the chart keeps deploying the embedded database pods, and the Appsmith pod's init containers gate startup on those unused internal databases — the init containers check only the `*.enabled` flags, not the external URLs.

Changes per guide (Kubernetes snippets):

- **MongoDB**: add `mongodb.enabled: false` and `mongodbCommunity.enabled: false` (Bitnami subchart and MongoDB Kubernetes Operator resource — whichever the installation uses).
- **PostgreSQL (RDS)**: add `postgresql.enabled: false` and `APPSMITH_KEYCLOAK_DB_DRIVER: postgresql`. Without the explicit driver, charts up to 3.9.0 resolve the driver to `h2` and Keycloak silently persists SSO data to a local file inside the pod instead of the external database (chart 3.9.1 infers it via appsmithorg/appsmith#42076, but the explicit value is correct on every version).
- **Redis**: add `redis.enabled: false`.


salevine (commented):
Looks good

salevine (approved):

PR #3034, opened 2026-08-19
docs: clarify Azure ACI storage guidance

sebastianiv21:
## Summary

- Move and simplify the Azure ACI best-practices section.
- Clarify that Azure Files persists non-database artifacts while MongoDB and Redis run externally.
- Link to the existing external MongoDB, Redis, and PostgreSQL setup guides.
- Pass database URLs and encryption values through secure ACI environment variables.

## Verification

- npm run build
- Verified all three internal guide targets.
- Verified the final section order and clean Markdown diff.

The build succeeds with the repository existing broken-anchor warnings; none originate from the Azure ACI page.

### DIFF
--- a/website/docs/getting-started/setup/installation-guides/azure-aci.md
+++ b/website/docs/getting-started/setup/installation-guides/azure-aci.md
@@ -6,7 +6,7 @@ toc_max_heading_level: 2
 
 # Azure Container Instance
 
-Azure Container Instances (ACI) is a simple and efficient way to run containers in the cloud. This document guides you through launching an ACI instance and running an Appsmith container. An Azure file share persists the files stored under `/appsmith-stacks`.
+Azure Container Instances (ACI) is a simple and efficient way to run containers in the cloud. This document guides you through launching an ACI instance and running an Appsmith container.
 
 :::note
 Azure only supports CIFS file shares and doesn't support NFS file shares.
@@ -14,7 +14,7 @@ Azure only supports CIFS file shares and doesn't support NFS file shares.
 
 ## Best practices
 
-For production deployments, use Azure Files only for Appsmith's filesystem artifacts. Don't use it to host databases or cache services because network-mounted volumes can cause poor performance, data inconsistency, or startup failures.
+For production deployments, use Azure Files only to persist Appsmith's filesystem artifacts. Don't use it to host databases or cache services because network-mounted volumes can cause poor performance, data inconsistency, or startup failures.
 
 Before deploying Appsmith, follow the relevant guides to configure the external services you need:
 
--- a/website/docs/getting-started/setup/installation-guides/azure-aci.md
+++ b/website/docs/getting-started/setup/installation-guides/azure-aci.md
@@ -44,18 +44,18 @@ fileShareName="myFileShareName"
 dnsNameLabel="myDNSLabel"
 ```
 
+Set the following variables using connection strings from the external service guides and secrets from your preferred secret-management method. Don't commit or share the actual values.
+
+```bash
+mongodbUrl="<mongodb-connection-url>"
+redisUrl="<redis-connection-url>"
+encryptionPassword="<strong-encryption-password>"
+encryptionSalt="<strong-encryption-salt>"
+```
+
 :::caution
 Use strong values for the encryption password and salt. Store them securely and reuse the same values whenever you redeploy or restore the instance. Changing them prevents Appsmith from decrypting stored credentials.
 :::
-
-Enter the MongoDB URL, Redis URL, encryption password, and encryption salt through silent prompts so the values aren't saved in your shell history:
-
-```bash
-read -rsp "MongoDB connection URL: " mongodbUrl && printf '\n'
-read -rsp "Redis connection URL: " redisUrl && printf '\n'
-read -rsp "Appsmith encryption password: " encryptionPassword && printf '\n'
-read -rsp "Appsmith encryption salt: " encryptionSalt && printf '\n'
-```
 
 ### Create a resource group (optional)
 
--- a/website/docs/getting-started/setup/installation-guides/azure-aci.md
+++ b/website/docs/getting-started/setup/installation-guides/azure-aci.md
@@ -9,11 +9,24 @@ import TabItem from '@theme/TabItem';
 
 # Azure Container Instance
 
-Azure Container Instances (ACI) is a simple and efficient way to run containers in the cloud. This document guides you through launching an ACI instance and running an Appsmith container. The data persists in Azure Storage Account File Share.
+Azure Container Instances (ACI) is a simple and efficient way to run containers in the cloud. This document guides you through launching an ACI instance and running an Appsmith container. An Azure file share persists configuration, certificates, Git data, logs, and other filesystem artifacts. External MongoDB and Redis services store application data, sessions, and cached data.
 
 :::note
 Azure only supports CIFS file shares and doesn't support NFS file shares.
 :::
+
+## Best practices
+
+For production deployments, don't run Appsmith's embedded MongoDB or Redis services on Azure Files. Database workloads are sensitive to the latency and filesystem behavior of network-mounted volumes, which can cause poor performance, data inconsistency, or startup failures.
+
+Use the following storage model:
+
+- Mount Azure Files at `/appsmith-stacks` to persist configuration, certificates, Git data, logs, and similar filesystem artifacts.
+- Use an [external MongoDB instance](/getting-started/setup/instance-configuration/custom-mongodb-redis) for Appsmith application data.
+- Use an [external Redis instance](/getting-started/setup/instance-configuration/external-redis) for sessions and caching.
+- Use [external PostgreSQL](/getting-started/setup/instance-configuration/external-postgresql-rds) when you enable SAML SSO or Workflows.
+
+The deployment command on this page follows this model by passing the MongoDB and Redis connection URLs as secure environment variables.
 
 ## Prerequisites​
 
@@ -21,11 +34,13 @@ Before launching an ACI instance, you need to have an Azure subscription and hav
 
 - [Azure Subscription](https://azure.com/free) - If you don't have an Azure subscription, you can sign up for a free trial.
 - [Azure CLI](https://learn.microsoft.com/en-us/cli/azure).
+- An authenticated [external MongoDB instance](/getting-started/setup/instance-configuration/custom-mongodb-redis) that the container can access. Configure TLS if your provider supports it.
+- An authenticated [external Redis instance](/getting-started/setup/instance-configuration/external-redis) that the container can access. Use the `rediss://` scheme when encryption in transit is enabled.
 - Whitelist `cs.appsmith.com` in your firewall settings to allow outbound HTTPS traffic. If using Azure Firewall, add these domains under Application Rules.
 
 ## Configure variables
 
-Update the following values starting with 'my' where necessary, and enter them in your shell/terminal.
+Update the following values for your deployment, and enter them in your shell or terminal.
 
 ```bash
 resourceGroupName="myResourceGroup"
@@ -34,6 +49,10 @@ storageAccountName="mystorageaccount$RANDOM"
 aciLocation="southindia"
 fileShareName="myFileShareName"
 dnsNameLabel="myDNSLabel"
+mongodbUrl="mongodb+srv://<username>:<password>@<host>/<database>"
+redisUrl="rediss://:<password>@<host>:<port>"
+encryptionPassword="<strong-random-password>"
+encryptionSalt="<long-random-string>"
 ```
 
 ### Create a resource group (optional)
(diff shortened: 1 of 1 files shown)


website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -34,6 +49,10 @@ storageAccountName="mystorageaccount$RANDOM"
 aciLocation="southindia"
 fileShareName="myFileShareName"
 dnsNameLabel="myDNSLabel"
+mongodbUrl="mongodb+srv://<username>:<password>@<host>/<database>"
+redisUrl="rediss://:<password>@<host>:<port>"
+encryptionPassword="<strong-random-password>"
```
sebastianiv21: **Encryption vars lack a must-match/preserve warning.** `APPSMITH_ENCRYPTION_PASSWORD`/`APPSMITH_ENCRYPTION_SALT` are presented as throwaway placeholders. An existing instance already has these values in `/appsmith-stacks/configuration/docker.env`; passing fresh random ones makes stored datasource credentials undecryptable, and a new user who doesn't record them can never restore a backup. `upgrade-migration-errors.md` documents exactly this failure and requires the values to match — worth a one-line caution here (keep these values safe; they must stay identical across redeploys and match any backup you restore).

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -72,18 +91,23 @@ az storage share create --name $fileShareName --account-name $storageAccountName
...
+    --resource-group $resourceGroupName \
+    --name $aciName \
+    --image appsmith/appsmith-ee:<version> \
+    --ip-address public \
+    --dns-name-label $dnsNameLabel \
+    --ports 80 443 \
+    --cpu 2 \
+    --memory 4 \
+    --secure-environment-variables \
+      "APPSMITH_DB_URL=$mongodbUrl" \
```
sebastianiv21, line 105: **No migration note for existing embedded-DB deployments.** The old command on this page ran embedded MongoDB with data on the file share. An existing ACI user re-running the new command comes up pointed at an empty external MongoDB — all workspaces/apps appear gone while the real data sits unreferenced on the share. Suggest a note pointing existing deployments to the *Restore MongoDB data* steps in `/getting-started/setup/instance-configuration/custom-mongodb-redis` before switching.

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -9,23 +9,38 @@ import TabItem from '@theme/TabItem';
...
+## Best practices
+
+For production deployments, don't run Appsmith's embedded MongoDB or Redis services on Azure Files. Database workloads are sensitive to the latency and filesystem behavior of network-mounted volumes, which can cause poor performance, data inconsistency, or startup failures.
+
+Use the following storage model:
+
+- Mount Azure Files at `/appsmith-stacks` to persist configuration, certificates, Git data, logs, and similar filesystem artifacts.
+- Use an [external MongoDB instance](/getting-started/setup/instance-configuration/custom-mongodb-redis) for Appsmith application data.
+- Use an [external Redis instance](/getting-started/setup/instance-configuration/external-redis) for sessions and caching.
+- Use [external PostgreSQL](/getting-started/setup/instance-configuration/external-postgresql-rds) when you enable SAML SSO or Workflows.
```
sebastianiv21: **The PostgreSQL guidance isn't backed by the command.** This bullet mandates external PostgreSQL for SAML SSO/Workflows, but the `az container create` block sets no `APPSMITH_KEYCLOAK_DB_URL` and *Configure variables* has no `postgresUrl` — while line 29 says the command "follows this model". A reader who deploys as written and then follows the *Configure SSO* card ends up with embedded Keycloak PostgreSQL initializing on the CIFS-mounted share — the exact anti-pattern this section warns about. Either add the variable to the command (as done for Mongo/Redis) or state explicitly that it must be added when enabling SSO/Workflows.

sebastianiv21: **Link target is AWS-specific.** `/getting-started/setup/instance-configuration/external-postgresql-rds` is titled *Configure External PostgreSQL (AWS RDS)* and its body is RDS-console steps (`rds.force_ssl`, RDS endpoints). The deleted section mapped this to **Azure Database for PostgreSQL** (and MongoDB Atlas in-region / Azure Cache for Redis); those Azure-native mappings were the reason this guidance lived on the Azure page. Consider restoring at least the Azure service names next to each bullet.

sebastianiv21 (commented):
Reviewed the rewrite — overall a clear improvement over the old *Best Practices & Things Not to Do* section: links resolve, env var names match the linked guides, and no inbound links break. The inline comments below cover 10 findings. The common thread in the top ones: the page now makes external MongoDB/Redis mandatory, but doesn't carry the safety rails that shift needs (encryption-key preservation, migration path for existing embedded-DB deployments, and the PostgreSQL half of the story).

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -34,6 +49,10 @@ storageAccountName="mystorageaccount$RANDOM"
 aciLocation="southindia"
 fileShareName="myFileShareName"
 dnsNameLabel="myDNSLabel"
+mongodbUrl="mongodb+srv://<username>:<password>@<host>/<database>"
+redisUrl="rediss://:<password>@<host>:<port>"
```
sebastianiv21: **Missing Azure Cache port guidance.** The generic `rediss://:<password>@<host>:<port>` placeholder invites the default port 6379, but Azure Cache for Redis — the natural choice on this page — serves TLS only on **6380** with the access key (documented in `external-redis.mdx`). Suggest using the canonical Azure form here: `rediss://:<access-key>@<name>.redis.cache.windows.net:6380`.

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -72,18 +91,23 @@ az storage share create --name $fileShareName --account-name $storageAccountName
...
+    --memory 4 \
+    --secure-environment-variables \
+      "APPSMITH_DB_URL=$mongodbUrl" \
+      "APPSMITH_REDIS_URL=$redisUrl" \
+      "APPSMITH_ENCRYPTION_PASSWORD=$encryptionPassword" \
+      "APPSMITH_ENCRYPTION_SALT=$encryptionSalt" \
+    --azure-file-volume-account-name $storageAccountName \
+    --azure-file-volume-account-key $storageAccountKey \
+    --azure-file-volume-share-name $fileShareName \
+    --azure-file-volume-mount-path "/appsmith-stacks/" \
```
sebastianiv21: **Dangling trailing backslash.** This is the last line of the block, so a copy-pasted command never executes — the shell sits at a continuation prompt, and the next thing typed is swallowed as an extra argument to `az container create`. Since this line was rewritten anyway, drop the trailing `\`.

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -9,23 +9,38 @@ import TabItem from '@theme/TabItem';
...
+
+The deployment command on this page follows this model by passing the MongoDB and Redis connection URLs as secure environment variables.
+
 ## Prerequisites​
 
 Before launching an ACI instance, you need to have an Azure subscription and have the Azure CLI installed on your machine.
 
 - [Azure Subscription](https://azure.com/free) - If you don't have an Azure subscription, you can sign up for a free trial.
 - [Azure CLI](https://learn.microsoft.com/en-us/cli/azure).
+- An authenticated [external MongoDB instance](/getting-started/setup/instance-configuration/custom-mongodb-redis) that the container can access. Configure TLS if your provider supports it.
```
sebastianiv21: **Prerequisite understates the linked guide's hard requirements.** `custom-mongodb-redis` requires MongoDB **5.0+**, an initialized **replica set**, and a user with `readWrite` + `clusterMonitor`; none of that is surfaced here (the Fargate guide states the 5.0 floor inline). Also, line 52 hardcodes `mongodb+srv://` — SRV rejects explicit ports, so self-hosted `host:port` users need the plain `mongodb://` form — and there's no URL-encoding caveat for passwords with special characters. Failures from any of these show up as a crash-looping container with nothing pointing back to the precondition.

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -9,23 +9,38 @@ import TabItem from '@theme/TabItem';
...
+- An authenticated [external Redis instance](/getting-started/setup/instance-configuration/external-redis) that the container can access. Use the `rediss://` scheme when encryption in transit is enabled.
 - Whitelist `cs.appsmith.com` in your firewall settings to allow outbound HTTPS traffic. If using Azure Firewall, add these domains under Application Rules.
 
 ## Configure variables
 
-Update the following values starting with 'my' where necessary, and enter them in your shell/terminal.
+Update the following values for your deployment, and enter them in your shell or terminal.
```
sebastianiv21, line 36: **Secrets pasted into an interactive shell.** This block now carries four production secrets (Mongo URL, Redis URL, encryption password/salt) that land in `~/.zsh_history` — undermining the `--secure-environment-variables` protection the command uses. The old block held only non-sensitive names. Worth a brief note (e.g. `read -s`, Key Vault, or history hygiene) for production use.

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -9,23 +9,38 @@ import TabItem from '@theme/TabItem';
...
 # Azure Container Instance
 
-Azure Container Instances (ACI) is a simple and efficient way to run containers in the cloud. This document guides you through launching an ACI instance and running an Appsmith container. The data persists in Azure Storage Account File Share.
+Azure Container Instances (ACI) is a simple and efficient way to run containers in the cloud. This document guides you through launching an ACI instance and running an Appsmith container. An Azure file share persists configuration, certificates, Git data, logs, and other filesystem artifacts. External MongoDB and Redis services store application data, sessions, and cached data.
 
 :::note
 Azure only supports CIFS file shares and doesn't support NFS file shares.
 :::
 
+## Best practices
```
sebastianiv21, line 15: **Structure diverges from every sibling install guide.** No other guide has a top-of-page *Best practices* section; Fargate handles the identical constraint as a single Prerequisites bullet, and Docker/AWS AMI/DigitalOcean keep the simple embedded-DB flow with externalization on the config pages. This page is now the only volume-backed guide that mandates external DBs in its primary command, with no lighter evaluation path — even though this section scopes its own rationale to "production deployments". Consider the sibling pattern: a Prerequisites bullet plus a `:::caution` beside the existing CIFS note.

website/docs/getting-started/setup/installation-guides/azure-aci.md
```diff
@@ -9,23 +9,38 @@ import TabItem from '@theme/TabItem';
...
+For production deployments, don't run Appsmith's embedded MongoDB or Redis services on Azure Files. Database workloads are sensitive to the latency and filesystem behavior of network-mounted volumes, which can cause poor performance, data inconsistency, or startup failures.
+
+Use the following storage model:
+
+- Mount Azure Files at `/appsmith-stacks` to persist configuration, certificates, Git data, logs, and similar filesystem artifacts.
+- Use an [external MongoDB instance](/getting-started/setup/instance-configuration/custom-mongodb-redis) for Appsmith application data.
+- Use an [external Redis instance](/getting-started/setup/instance-configuration/external-redis) for sessions and caching.
+- Use [external PostgreSQL](/getting-started/setup/instance-configuration/external-postgresql-rds) when you enable SAML SSO or Workflows.
+
+The deployment command on this page follows this model by passing the MongoDB and Redis connection URLs as secure environment variables.
```
sebastianiv21: **Redundancy within the page.** The external-Mongo/Redis guidance appears three times (intro, this section, Prerequisites) with the same two links duplicated; the artifact list on line 24 near-verbatim repeats line 12; this sentence narrates what the command block already shows; and the `Tabs`/`TabItem` imports (lines 7–8) are unused in the rewritten file. Suggest stating the requirement once in Prerequisites, dropping this sentence, and removing the dead imports.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Added a caution requiring strong encryption values to be stored securely and reused across redeployments and restores.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Omitted intentionally because this page covers new deployments. Migration guidance remains in the dedicated external MongoDB documentation.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Removed the claim that the base command configures PostgreSQL. The page now directs readers to the dedicated SAML and Workflow prerequisites before enabling those features.

sebastianiv21: Replaced the AWS-specific recommendation with the Azure SAML PostgreSQL guide and the dedicated Workflow prerequisites.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Removed the generic Redis URI example. Provider-specific connection details remain in the external Redis guide.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Removed the trailing continuation character and validated the combined Bash snippets with bash -n.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Removed the hardcoded MongoDB URI and kept the detailed requirements in the linked external MongoDB guide as the source of truth.

sebastianiv21: Removed the hardcoded MongoDB URI and kept the detailed requirements in the linked external MongoDB guide as the source of truth.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Replaced plaintext secret assignments with read -rsp prompts so values are not recorded in shell history. The ACI command still uses secure environment variables.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Kept the Best practices section intentionally because moving and simplifying it is the goal of this PR. The section is now concise and the duplicated guidance has been removed.

website/docs/getting-started/setup/installation-guides/azure-aci.md
sebastianiv21: Removed the unused imports, repeated artifact and service descriptions, duplicated guide links, and command narration.

salevine (approved):
Let

PR #3036, opened 2026-08-21
fix: replace stale managed vector store files

sebastianiv21:
## Description

- tag newly attached documentation files with a workflow ownership attribute
- delete older files only when they carry the same workflow ownership attribute
- preserve all unmanaged files and files managed by another source
- wait for successful ingestion before cleanup and verify the resulting store contents
- bound API requests, pagination, verification retries, and cleanup runtime
- add mocked regression coverage that runs before OpenAI mutations

## Test plan

- [x] Validate workflow YAML and step structure
- [x] Validate Bash syntax
- [x] Run six mocked cleanup scenarios covering pagination, delayed removal, unmanaged-file preservation, deletion failure, repeated cursors, page limits, and malformed responses

Result: 6 cleanup tests passed.

### DIFF
--- a/.github/scripts/cleanup-openai-vector-store.sh
+++ b/.github/scripts/cleanup-openai-vector-store.sh
@@ -1,0 +1,254 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${OPENAI_API_KEY:?OPENAI_API_KEY is required}"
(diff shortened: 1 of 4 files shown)


salevine (commented):
LGTM

salevine (approved):

PR #3037, opened 2026-09-01
docs: point ECS health checks at /api/v1/health

sebastianiv21:
## Description

The ECS installation guides (Fargate and EC2) configure both the container-level health check and the ALB target-group health checks against `/`. That path is served by Caddy and returns 200 even when the backend has lost database connectivity, so ECS keeps a degraded task in service.

This changes all five health-check references to `/api/v1/health`, which verifies MongoDB and Redis connectivity, so ECS replaces a task whose DB connection has gone stale. It matches what the troubleshooting guide already recommends for load balancer probes.

EC2 guide additionally:
- `curl` → `curl -f`: without `-f`, curl exits 0 on a 5xx response, so the check could never fail on HTTP errors.
- Health check timeout 2s → 5s, matching the Fargate guide, since the endpoint performs real DB pings.

## Context

Pylon issue 2854: an ECS deployment's Mongo driver held a stale replica-set topology (no primary in the client view) and the task stayed in service until manually restarted, because the `/` health check kept returning 200 throughout the outage.

## Behavior notes

- If the external database is down for an extended period, tasks now restart-loop instead of sitting degraded — intended for this failure class.
- The container check now requires the Java backend to be up; the documented start period (160s) plus retries gives ~190s before a booting task is killed, unchanged from before.


salevine (approved):
LGTM

PR #3038, opened 2026-09-02
fix: detach stale files from vector store before deleting them

sebastianiv21:
## Summary

Fixes the "Upload Docs to OpenAI Vector Store" workflow failing in the "Remove stale workflow-managed files" step ([failed run](https://github.com/appsmithorg/appsmith-docs/actions/runs/33529572639/job/99928974488)).

The cleanup script deleted stale files through the Files API only (`DELETE /v1/files/{id}`). The vector store kept listing the deleted entry, so the verification loop exhausted all six attempts and the job exited 1.

## Changes

- `cleanup-openai-vector-store.sh`: detach each stale file from the vector store (`DELETE /v1/vector_stores/{vs}/files/{id}`) before deleting the file object. Log an explicit `ORPHANED FILE` warning if the file delete fails after a successful detach, since later runs cannot rediscover a file that is no longer attached.
- Test mock (`fixtures/curl`): add the detach route, an ordering guard that fails if a file is deleted before being detached, and a `detach-failure` scenario.
- Test suite: assert detach markers in the success scenario and add a detach-failure case that confirms the script stops before deleting the file object.

## Testing

```
bash .github/scripts/tests/cleanup-openai-vector-store.test.sh
7 cleanup tests passed
```

Shellcheck was not run locally (not installed).

## Known risk

If a run is killed between the detach and the file delete, the file object is orphaned in OpenAI storage and no later run will find it. Low severity, now surfaced in the log. A periodic sweep of `GET /v1/files` would be a follow-up if this ever matters.

### DIFF
--- a/.github/scripts/cleanup-openai-vector-store.sh
+++ b/.github/scripts/cleanup-openai-vector-store.sh
@@ -184,9 +184,24 @@ if [ "${#stale_file_ids[@]}" -eq 0 ]; then
 fi
 
 for stale_file_id in ${stale_file_ids[@]+"${stale_file_ids[@]}"}; do
+  echo "Detaching stale file $stale_file_id from vector store..."
+  if ! detach_response=$(api_delete "$OPENAI_API_BASE_URL/vector_stores/$VECTOR_STORE_ID/files/$stale_file_id"); then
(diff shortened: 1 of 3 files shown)


salevine (approved):
Lgtm

